PR #112, opened 2026-04-22
feat: add minimal todo CRUD example mirroring src/user layout

### LINKED ISSUE #93: examples/todo: minimal CRUD example
Part of #85. First of 5 seed examples.

## What to build

A minimal `todo` domain under `examples/todo/` that mirrors the layout
of `src/user/` (without password hashing, admin, or worker).

## Acceptance criteria

- `examples/todo/domain/dtos/todo_dto.py` — `TodoDTO(id, title, description, done, created_at, updated_at)`
- `examples/todo/domain/protocols/todo_repository_protocol.py`
- `examples/todo/domain/services/todo_service.py` — inherits `BaseService[CreateTodoRequest, UpdateTodoRequest, TodoDTO]`
- `examples/todo/infrastructure/database/models/todo_model.py` (`__tablename__ = "todo"`)
- `examples/todo/infrastructure/repositories/todo_repository.py`
- `examples/todo/infrastructure/di/todo_container.py`
- `examples/todo/interface/server/schemas/todo_schema.py`
- `examples/todo/interface/server/routers/todo_router.py` — 5 standard endpoints: `POST /todo`, `GET /todos`, `GET /todo/{id}`, `PUT /todo/{id}`, `DELETE /todo/{id}`
- `examples/todo/interface/server/bootstrap/todo_bootstrap.py`
- `examples/todo/tests/test_todo_service.py` — protocol-based mock, ≥ 2 passing tests
- `examples/todo/README.md` — one-paragraph summary + 3 `curl` commands

## Verification

1. Copy `examples/todo/` → `src/todo/`
2. `rm -f ./quickstart.db && make quickstart`
3. In another terminal, exercise the 5 endpoints with `curl` (see your `examples/todo/README.md`)
4. `pytest examples/todo/tests/ -v` passes

## References

- Tutorial walk-through (the same pattern, with `order` instead of `todo`): https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/docs/tutorial/first-domain.md
- Production reference: https://github.com/Mr-DooSun/fastapi-agent-blueprint/tree/main/src/user
- Examples guide (landing in #92): https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/92

## Out of scope

Authentication, tags, soft delete, batch create. Keep it boring — this example exists so the *next* examples can reference it.

Naman9245:
## Related Issue
- Closes #93

## Change Summary
Added minimal todo CRUD example mirroring the src/user/ layout.

## Type of Change
- [x] feat: New feature

## Checklist
- [x] Tests pass
- [ ] Linting passes ('ruff check src/')

## How to Test
```bash
PYTHONPATH=. pytest examples/todo/tests/ -v
```
2 tests pass.

## Files Added
- Domain: todo_dto.py, todo_repository_protocol.py, todo_service.py
- Infrastructure: todo_model.py, todo_repository.py, todo_container.py
- Interface: todo_schema.py, todo_router.py, todo_bootstrap.py
- Tests: test_todo_service.py
- README.md with curl commands


examples/todo/domain/protocols/todo_repository_protocol.py
```diff
@@ -0,0 +1,18 @@
+from typing import Protocol, List, Optional
+from examples.todo.domain.dtos.todo_dto import TodoDTO
+
+class TodoRepositoryProtocol(Protocol):
```
Mr-DooSun: **[BLOCKING] Wrong inheritance — protocol is a dead abstraction**

`TodoRepositoryProtocol` inherits from `typing.Protocol` directly and defines methods (`create`, `get`, `list`, `update`, `delete`) that `BaseService` never calls. `BaseService` internally uses `insert_data`, `select_data_by_id`, `update_data_by_data_id`, `delete_data_by_data_id` — none of which exist here.

Reference pattern from `UserRepositoryProtocol`:
```python
from src._core.domain.protocols.repository_protocol import BaseRepositoryProtocol
from examples.todo.domain.dtos.todo_dto import TodoDTO

class TodoRepositoryProtocol(BaseRepositoryProtocol[TodoDTO]):
    pass
```

Mr-DooSun (changes requested):
**BLOCKING: 4 | SUGGESTION: 4 | CI failure: architecture job**

Thanks for the contribution — the layer structure (3-TypeVar `BaseService`, DI container wiring, ADR 004 schema import) is correctly applied. However there are 4 runtime bugs the mocked tests don't catch, plus a CI failure. Details are in the inline comments below.

examples/todo/interface/server/routers/todo_router.py
```diff
@@ -0,0 +1,38 @@
...
+@inject
+async def create_todo(item: CreateTodoRequest, todo_service: TodoService = Depends(Provide[TodoContainer.todo_service])) -> SuccessResponse[TodoResponse]:
+    data = await todo_service.create_data(entity=item)
+    return SuccessResponse(data=TodoResponse(**data.model_dump()))
+
+@router.get("/todos", summary="List todos", response_model=SuccessResponse[list[TodoResponse]], response_model_exclude={"pagination"})
+@inject
+async def list_todos(todo_service: TodoService = Depends(Provide[TodoContainer.todo_service])) -> SuccessResponse[list[TodoResponse]]:
+    data = await todo_service.get_data_list()
+    return SuccessResponse(data=[TodoResponse(**d.model_dump()) for d in data])
```
Mr-DooSun: **[BLOCKING] `get_data_list()` does not exist in `BaseService` → `AttributeError` at runtime**

`BaseService` has no `get_data_list()` method. `GET /todos` will always return 500.

The correct method is:
```python
datas, pagination = await todo_service.get_datas(page=1, page_size=100)
return SuccessResponse(data=[TodoResponse(**d.model_dump()) for d in datas])
```

examples/todo/infrastructure/database/models/todo_model.py
```diff
@@ -0,0 +1,13 @@
+from datetime import datetime
+from sqlalchemy import Boolean, Column, DateTime, String, Text
+from src._core.infrastructure.database.database import Base
```
Mr-DooSun: **[BLOCKING] Import path does not exist → `ModuleNotFoundError`**

`src/_core/infrastructure/database/` does not exist in this project.

```python
# Wrong
from src._core.infrastructure.database.database import Base

# Correct
from src._core.infrastructure.persistence.rdb.database import Base
```

`todo_repository.py` and `todo_container.py` will fail to load. The 2 tests passed only because they mock the repository and never import this file.

examples/todo/infrastructure/database/models/todo_model.py
```diff
@@ -0,0 +1,13 @@
...
+
+class TodoModel(Base):
+    __tablename__ = "todo"
+
+    id = Column(String(36), primary_key=True, index=True)
```
Mr-DooSun: **[BLOCKING] No default value for `id` → `INSERT` fails against a real DB**

`CreateTodoRequest` does not include `id`. `BaseRepository.insert_data` calls `TodoModel(**entity.model_dump(exclude_none=True))` without an `id` value → NOT NULL constraint violation on `POST /todo`.

Recommended fix — use an auto-increment integer PK to stay consistent with `UserModel` and `BaseRepository`'s `data_id: int` interface:
```python
from sqlalchemy import Integer
id = Column(Integer, primary_key=True, autoincrement=True)
```

If UUID strings are preferred, add a default and update `TodoDTO.id` and the router path param type accordingly.

tests/unit/todo/domain/test_todo_service.py
```diff
@@ -0,0 +1,36 @@
...
+    service = TodoService(todo_repository=mock_repo)
+    request = CreateTodoRequest(title="Test Todo", description="Test desc")
+    result = await service.create_data(entity=request)
+    assert result.title == "Test Todo"
+    assert result.done is False
+
+@pytest.mark.asyncio
+async def test_get_todo():
+    mock_repo = AsyncMock()
+    mock_repo.select_data_by_id.return_value = make_todo_dto(title="Fetched Todo")
```
Mr-DooSun, line 41: **[CI FAILURE] ruff S101 — `assert` in `examples/` is not covered by the existing `per-file-ignores`**

The project's ruff config has `per-file-ignores` covering `tests/**` but not `examples/**/tests/`. This causes the `architecture` CI job to fail.

Fix options:
1. Extend `per-file-ignores` in `pyproject.toml` or `ruff.toml` to include `"examples/**/tests/**": ["S101"]`
2. Move tests to the standard `tests/unit/` path

Also note: 5 files need `ruff format` (missing newlines at end of file). Run `ruff format examples/` locally before pushing.

examples/todo/infrastructure/database/models/todo_model.py
```diff
@@ -0,0 +1,13 @@
...
+    title = Column(String(255), nullable=False)
+    description = Column(Text, nullable=True)
+    done = Column(Boolean, default=False)
+    created_at = Column(DateTime, default=datetime.utcnow)
```
Mr-DooSun: **[SUGGESTION] `datetime.utcnow` is deprecated since Python 3.12**

```python
# Before
default=datetime.utcnow, onupdate=datetime.utcnow

# After
from datetime import datetime, timezone
default=lambda: datetime.now(timezone.utc)
onupdate=lambda: datetime.now(timezone.utc)
```

Note: the test file already uses `datetime.now(UTC)` correctly.

examples/todo/interface/server/bootstrap/todo_bootstrap.py
```diff
@@ -0,0 +1,20 @@
...
+def create_todo_container(todo_container: TodoContainer):
+    todo_container.wire(packages=["examples.todo.interface.server.routers"])
+    return todo_container
+
+def setup_todo_routes(app: FastAPI):
+    """Register todo domain routes"""
+    app.include_router(router=todo_router.router, prefix="/v1", tags=["Todo"])
+
+def bootstrap_todo_domain(
+    app: FastAPI, database: Database, todo_container: TodoContainer
```
Mr-DooSun: **[SUGGESTION] Unused `database` parameter**

`database` is received but never used in the function body. The container already gets `database` via `core_container.database` through DI.

Reference `user_bootstrap.py` signature:
```python
def bootstrap_user_domain(app: FastAPI, user_container: UserContainer) -> None:
```

Naman9245:
Hi @Mr-DooSun,

Thank you so much for the detailed review and guidance! I have pushed a new commit addressing all of your feedback:

Fixed the BaseRepositoryProtocol inheritance in todo_repository_protocol.py.

Updated the list endpoint in the router to correctly use get_data and handle pagination.

Changed the id field from String to Integer across the Model, DTO, and Schema files.

Removed the database parameter from the bootstrap signature.

I have run both pytest and ruff check --fix locally, and all tests are passing with clean formatting.

Let me know if there is anything else you need me to adjust. Thanks again for helping me with my first open-source contribution!

Naman9245:
Hi @Mr-DooSun, just a quick follow-up! I also just pushed one more commit to clear up the CI failure you mentioned:

Ran ruff format examples/ to fix the missing newlines at the end of the files.

Moved the tests to the standard tests/unit/todo/ path so the existing ruff config correctly ignores the S101 assert warnings.

Both pytest tests/unit/todo/ -v and ruff check are completely green locally now. Thanks again!

examples/todo/interface/server/routers/todo_router.py
```diff
@@ -0,0 +1,86 @@
...
+    return SuccessResponse(data=TodoResponse(**data.model_dump()))
+
+
+@router.get(
+    "/todos",
+    summary="List todos",
+    response_model=SuccessResponse[list[TodoResponse]],
+    response_model_exclude={"pagination"},
+)
+@inject
```
Mr-DooSun, line 36: **[BLOCKING] Still calling a non-existent method**

You renamed `get_data_list` → `get_data`, but `BaseService` exposes `get_datas` (plural). The previous review comment showed the correct call:

```python
datas, pagination = await todo_service.get_datas(page=1, page_size=100)
```

Reference: `src/_core/domain/services/base_service.py:29` — the real signature is `async def get_datas(...)`. `get_data` / `get_data_list` don't exist, so `GET /todos` will still raise `AttributeError` → 500.

The reason this wasn't caught locally is that `tests/unit/todo/test_todo_service.py` uses `AsyncMock`, so any method name is silently accepted by the mock.

examples/todo/README.md
```diff
@@ -0,0 +1,23 @@
+# Todo Example
```
Mr-DooSun, line 1: **[DESIGN] `examples/todo/` is not wired into the running server**

`discover_domains()` in `src/_core/infrastructure/discovery.py:21` scans `src/` only, and `src/_apps/server/app.py` does not import `examples/todo/` bootstrap. So even after the BLOCKING fix, none of the endpoints listed in this README actually exist on the live server.

Three options — I'd like your (and @Mr-DooSun's) opinion before merge:

1. **Move to `src/todo/`** — becomes a real domain, auto-discovered, endpoints work out of the box. Simplest, but adds a permanent domain to the repo.
2. **Keep under `examples/` and document clearly** — add a prominent note to this README: *"Reference code only. Not auto-wired. To run it, copy into `src/todo/` or add a manual bootstrap call."* This preserves the "example" intent and keeps the main tree clean.
3. **Extend discovery to scan `examples/`** — ❌ not recommended; risks example code leaking into production boot.

My preference is **(2)** if the goal is a read-only reference for new contributors, or **(1)** if you want it to be a working demo that can be curl'd. Either way, this PR is the first to establish the convention, so documenting the decision in `AGENTS.md` / `docs/ai/shared/project-dna.md` should follow (`/sync-guidelines`).

tests/unit/todo/domain/test_todo_service.py
```diff
@@ -0,0 +1,44 @@
...
+    assert result.done is False
+
+
+@pytest.mark.asyncio
+async def test_get_todo():
+    mock_repo = AsyncMock()
+    mock_repo.select_data_by_id.return_value = make_todo_dto(title="Fetched Todo")
+    service = TodoService(todo_repository=mock_repo)
+    result = await service.get_data_by_data_id(data_id=123)
+    assert result.title == "Fetched Todo"
```
Mr-DooSun, line 44: **[NOTE] Mock-only test coverage hides runtime bugs**

`AsyncMock` silently accepts any method name, which is why `get_data_list` (original PR) and now `get_data` (current fix) both slipped through. The previous review called this out, and the same pattern recurred.

If this PR is meant to be a template others copy via `/new-domain`, consider adding a tiny integration test — even a single `tests/e2e/todo/test_todo_e2e.py` that boots the app with SQLite in-memory and hits `POST /v1/todo` + `GET /v1/todos` — to catch service method naming errors automatically. Not blocking for this PR, just a suggestion for the template's long-term value.

Mr-DooSun (changes requested):
## Re-review summary

Thanks for the quick turnaround — 6 of the original 7 findings are resolved and CI is green. I'm resolving those threads.

However there is **1 BLOCKING** issue still present and **1 HIGH** design question worth discussing before merge:

- **[BLOCKING]** `todo_router.py:37` still calls a non-existent method. You renamed `get_data_list` → `get_data`, but `BaseService` has `get_datas` (plural — see `src/_core/domain/services/base_service.py:29`). Runtime `AttributeError` remains. See inline comment.
- **[HIGH / DESIGN]** `examples/todo/` is not picked up by `discover_domains()` (which scans `src/` only), and there is no manual wiring in `src/_apps/server/app.py`. The README advertises live endpoints (`POST /v1/todo`, …) but the server never mounts them. This is also why the `get_data` bug survived the re-review — the mock-only tests don't exercise the real router. See inline comment on the README with options.

### Note on testing
The mock-only test pattern hid this exact kind of runtime bug twice now. If this PR is meant to be a template others copy, adding a single integration-level test (spinning up the FastAPI app and hitting `/v1/todos`) would catch the `get_datas` / `get_data_by_data_id` naming issues automatically. Not required for merge — just a suggestion for the template's long-term value.

Once the BLOCKING is fixed and we agree on the `examples/` design decision, this is ready.

examples/todo/infrastructure/database/models/todo_model.py
```diff
@@ -0,0 +1,18 @@
+from datetime import UTC, datetime, timezone
+
+from sqlalchemy import Boolean, Column, DateTime, Integer, String, Text
+
+from src._core.infrastructure.persistence.rdb.database import Base
+
+
+class TodoModel(Base):
+    __tablename__ = "todo"
```
Mr-DooSun, line 10: **[MEDIUM] SQLAlchemy 1.x style — doesn't actually mirror `UserModel`**

The PR title says *"mirroring src/user layout"*, but this Model is written in SQLAlchemy 1.x style (`Column(...)`) while `UserModel` uses the 2.0 declarative style (`Mapped[...] / mapped_column(...)`) which is the project convention (`AGENTS.md`, `project-dna`).

`src/user/infrastructure/database/models/user_model.py`:
```python
from sqlalchemy import DateTime, Integer, String, func
from sqlalchemy.orm import Mapped, mapped_column

class UserModel(Base):
    __tablename__ = "user"

    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
    username: Mapped[str] = mapped_column(String(20), nullable=False)
    created_at: Mapped[DateTime] = mapped_column(
        DateTime, server_default=func.now(), nullable=True
    )
    updated_at: Mapped[DateTime] = mapped_column(
        DateTime, server_default=func.now(), onupdate=func.now(), nullable=True
    )
```

Porting `TodoModel` to the same style also lets you drop the Python-level `lambda: datetime.now(UTC)` defaults in favor of `server_default=func.now()`, which is what user uses.

This matters because: if this PR becomes a template copy-pasted by other contributors, the 1.x style propagates.

examples/todo/interface/server/routers/todo_router.py
Mr-DooSun, line 36: **[MEDIUM] Hardcoded pagination + dropped `pagination` response — doesn't mirror user router either**

Separate from the `get_data` → `get_datas` BLOCKING issue above, the whole endpoint signature diverges from `user_router.get_user`:

```python
# src/user/interface/server/routers/user_router.py
async def get_user(
    page: int = Query(1, ge=1),
    page_size: int = Query(10, alias="pageSize", ge=1, le=100),
    user_service: UserService = Depends(Provide[UserContainer.user_service]),
) -> SuccessResponse[list[UserResponse]]:
    datas, pagination = await user_service.get_datas(page=page, page_size=page_size)
    return SuccessResponse(
        data=[UserResponse(**data.model_dump(exclude={"password"})) for data in datas],
        pagination=pagination,
    )
```

Current Todo version has no `Query` parameters (client can't paginate), hardcodes `page=1, page_size=100`, and throws away `pagination` in the response (also `response_model_exclude={"pagination"}` drops it at schema level). For a "minimal example mirroring user", this is exactly the place the mirror should hold — paginated list is the canonical pattern learners will copy.

Recommended fix while you're already touching this line for the `get_datas` rename:
```python
async def list_todos(
    page: int = Query(1, ge=1),
    page_size: int = Query(10, alias="pageSize", ge=1, le=100),
    todo_service: TodoService = Depends(Provide[TodoContainer.todo_service]),
) -> SuccessResponse[list[TodoResponse]]:
    datas, pagination = await todo_service.get_datas(page=page, page_size=page_size)
    return SuccessResponse(
        data=[TodoResponse(**d.model_dump()) for d in datas],
        pagination=pagination,
    )
```

And drop `response_model_exclude={"pagination"}` from the `@router.get("/todos")` decorator.

Mr-DooSun (commented):
## Follow-up observations (inline)

Two more inline comments posted separately — see `todo_model.py` and `todo_router.py:37`. The general comment above covers C/D/E nits.

Mr-DooSun (commented):
## Follow-up observations

I did a deeper pass on the diff focused on one question: does this PR actually *mirror* `src/user/` as its title claims? Found a few gaps worth sharing. None of these block merge on their own, but A and B directly contradict the "mirroring src/user layout" framing — worth fixing since this PR is likely to be used as a template by future contributors.

Inline comments below cover:
- **A** — `TodoModel` uses SQLAlchemy 1.x style instead of the 2.0 `Mapped[...] / mapped_column(...)` style that `UserModel` uses.
- **B** — `list_todos` hardcodes pagination instead of exposing `Query` parameters, and drops the pagination response.

### Nits worth a quick sweep (C / D / E)

**C. `__init__.py` coverage is inconsistent.** `src/user/` has `__init__.py` in *every* subdirectory (even empty ones). `examples/todo/` has them in some (infrastructure/di, interface/server/...) but not in others (`examples/`, `examples/todo/`, `examples/todo/domain/`, `examples/todo/domain/dtos/`, `examples/todo/domain/protocols/`, `examples/todo/domain/services/`, `examples/todo/infrastructure/database/`, `examples/todo/infrastructure/database/models/`). Works due to PEP 420 namespace packages, but diverges from the project convention — and `/new-domain` scaffolds them everywhere.

**D. Unused imports.** CI passes because `pyproject.toml` has `F401` in `ignore`, but these are dead:
- `examples/todo/domain/dtos/todo_dto.py:2` — `from typing import Optional` (file uses `str | None`)
- `examples/todo/interface/server/schemas/todo_schema.py:2` — same
- `examples/todo/infrastructure/database/models/todo_model.py:1` — `from datetime import UTC, datetime, timezone` — `timezone` is unused

**E. Model columns missing explicit `nullable=False`.** `done`, `created_at`, `updated_at` don't specify `nullable`, so SQLAlchemy defaults to `nullable=True` on those columns. The DTO/Response expect non-null, so if a default ever fails to fire, DTO validation breaks. `UserModel` specifies `nullable=True` or `nullable=False` explicitly on every column — worth mirroring.

Mr-DooSun (commented):
## Branch is behind main — please rebase before addressing fixes

PR base is at `9074982` (merge of #109), but `main` is now at `c425839` — **12 commits ahead**. Physical file conflicts: **none** (the files this PR adds don't overlap with anything main changed). But two semantic changes on `main` affect this PR:

### 1. `user_bootstrap.py` was simplified (ADR 043, #111)

Your `todo_bootstrap.py` still mirrors the *old* user pattern. After rebase, please follow the new canonical shape:

```python
# Before (your current PR — mirrors old user_bootstrap)
def create_todo_container(todo_container: TodoContainer):
    todo_container.wire(packages=["examples.todo.interface.server.routers"])
    return todo_container

def bootstrap_todo_domain(app: FastAPI, todo_container: TodoContainer) -> None:
    todo_container = create_todo_container(todo_container=todo_container)
    setup_todo_routes(app=app)

# After (new user_bootstrap on main)
def create_todo_container(todo_container: TodoContainer) -> None:
    todo_container.wire(packages=["examples.todo.interface.server.routers"])

def bootstrap_todo_domain(app: FastAPI, todo_container: TodoContainer) -> None:
    create_todo_container(todo_container=todo_container)
    setup_todo_routes(app=app)
```

Changes: `create_*_container` no longer returns (return type is now `-> None`), and `bootstrap_*_domain` just calls it without re-assigning.

### 2. Canonical required test files were formalized (#113, `docs/ai/shared/test-files.md`)

Main now has an explicit contract for per-domain test files that `/review-architecture §5` enforces. Your PR was written before this landed, so it doesn't conform.

**Baseline (always required)**:

| File | Status in PR #112 |
|---|---|
| `tests/factories/todo_factory.py` | ❌ Missing |
| `tests/unit/todo/domain/test_todo_service.py` | ⚠ Present at `tests/unit/todo/test_todo_service.py` — needs a `domain/` subdirectory |
| `tests/integration/todo/infrastructure/test_todo_repository.py` | ❌ Missing |

**Conditional (required because this PR adds a server router)**:

| File | Status |
|---|---|
| `tests/e2e/todo/test_todo_router.py` | ❌ Missing |

If we agree this PR should mirror `/new-domain` scaffolding (which your PR title claims), the e2e test is especially important — it's the one that would have caught the `get_datas` BLOCKING bug that slipped through twice on mock-only unit tests. That's the canonical answer to my earlier mock-only concern.

See [`docs/ai/shared/test-files.md`](https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/docs/ai/shared/test-files.md) for the full spec.

### Suggested next-step order

1. `git rebase origin/main` (no conflicts expected)
2. Fix the BLOCKING `get_data` → `get_datas` in router
3. Resolve the `examples/` vs `src/todo/` placement question
4. Update `todo_bootstrap.py` to the simplified shape above
5. Restructure tests to match `docs/ai/shared/test-files.md` (move service test to `tests/unit/todo/domain/`, add factory + integration + e2e)
6. Apply the mirroring fixes from my earlier reviews (SQLAlchemy 2.0 style, `Query` pagination params, `__init__.py` coverage, unused imports, explicit `nullable=False`)

Happy to re-review once rebased.

Naman9245:
Hi @Mr-DooSun,

Thank you for the incredibly detailed re-review! I have rebased with main and addressed all the new standards and blocking issues:

Router (todo_router.py): Fixed the get_datas call and added the page and page_size Query parameters to perfectly mirror the user router pattern.

Database Model (todo_model.py): Upgraded entirely to the SQLAlchemy 2.0 Mapped / mapped_column style and explicitly set nullable=False where needed (Nit E).

Bootstrap (todo_bootstrap.py): Updated to match the new ADR 043 pattern (removed the database parameter, updated return types to None).

Test Structure: Created the domain, factories, and integration folders to match the new test layout contract, and moved test_todo_service.py to tests/unit/todo/domain/.

README & Init files: Added the design note (Option 2) to the top of the README, and added all the missing __init__.py files across the directories (Nit C).

Unused Imports: Ran ruff check --fix to clean up all dead imports (Nit D).

Both pytest and ruff are completely green locally against the new structure. Please take a look, and if there are any other issues or adjustments needed, just let me know! Thanks again for guiding me through this and helping me make this a solid template.

examples/todo/domain/dtos/todo_dto.py
```diff
@@ -0,0 +1,13 @@
+from datetime import datetime
+from typing import Optional
```
Mr-DooSun: **[NOTE] Unused import** — the file uses `str | None` throughout (PEP 604 syntax), so `Optional` is never referenced.

Why your `ruff check --fix` didn't catch this: `pyproject.toml` has `F401` in ruff's ignore list project-wide, so unused imports aren't flagged anywhere in the repo. Your claim was technically true (ruff was clean) but the intent (removing dead imports) didn't apply. Manual delete is the only path.

Same applies to `examples/todo/interface/server/schemas/todo_schema.py:2`.

examples/todo/interface/server/schemas/todo_schema.py
```diff
@@ -0,0 +1,27 @@
+from datetime import datetime
+from typing import Optional
```
Mr-DooSun: **[NOTE] Same as `todo_dto.py:2`** — unused `Optional`, not caught by `ruff check --fix` because `F401` is in the project's ruff ignore list.

Mr-DooSun (commented):
Hi @Naman9245,

Thanks for the quick turnaround on the rebase and the structural fixes. I re-read the diff one more time and realized two of my Round 2 asks were inconsistent with the Option 2 design you adopted — letting me correct those and scale the review back.

## Re-review — scaling back, 2 minor items left

### Retracting
- **`tests/e2e/todo/test_todo_router.py` requirement** — Option 2 means there is no FastAPI mount point, so an e2e test is impossible by design. My previous ask didn't account for the design you chose. Dropping it.
- **Full test-files baseline compliance** — `docs/ai/shared/test-files.md` targets `src/{name}/` domains (enforced by `/review-architecture §5`, which only scans `src/`). Applying it rigidly to `examples/` was over-strict on my part.

### Still worth fixing (both minor)
- **[LOW] Empty test placeholders** — `tests/factories/todo_factory.py` and `tests/integration/todo/infrastructure/test_todo_repository.py` are 0 bytes. **Either fill them with real tests, or just delete them** — deletion is a completely valid choice for a reference-only example (no obligation to demo the test layer).
- **[NOTE] Unused `Optional` imports** — see 2 inline comments. Your `ruff check --fix` didn't catch them because `pyproject.toml` has `F401` in ruff's ignore list project-wide (likely for `__init__.py` re-exports), so unused imports never get flagged. Manual delete is the only option.

### What landed well
- BLOCKING `get_datas` fix — verified against `BaseService.get_datas` in `src/_core/domain/services/base_service.py:29`
- SQLAlchemy 2.0 style — your `Mapped[datetime]` is actually **more correct** than `UserModel`, which uses `Mapped[DateTime]` (SQLAlchemy column type) where the Python `datetime` type belongs. Nice catch.
- Pagination `page` / `pageSize` mirrors `src/user/interface/server/routers/user_router.py:65` exactly
- Bootstrap matches the ADR 043 simplified shape
- Option 2 + README note — clean resolution of the discovery question

Once the two nits are addressed, this is good to merge. Thanks again for the iteration patience — this is a solid first OSS contribution.

examples/todo/infrastructure/database/models/todo_model.py
```diff
@@ -0,0 +1,17 @@
...
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    title: Mapped[str] = mapped_column(String(255), nullable=False)
+    description: Mapped[str | None] = mapped_column(Text, nullable=True)
+    done: Mapped[bool] = mapped_column(Boolean, default=False, nullable=False)
+    created_at: Mapped[datetime] = mapped_column(DateTime, server_default=func.now(), nullable=False)
+    updated_at: Mapped[datetime | None] = mapped_column(DateTime, server_default=func.now(), onupdate=func.now(), nullable=True)
```
Mr-DooSun: **[CI] `ruff format` wants these two lines wrapped** — both `created_at` (line 16) and `updated_at` (line 17) exceed the 88-char line limit, so `ruff format` reformats them like this:

```python
created_at: Mapped[datetime] = mapped_column(
    DateTime, server_default=func.now(), nullable=False
)
updated_at: Mapped[datetime | None] = mapped_column(
    DateTime, server_default=func.now(), onupdate=func.now(), nullable=True
)
```

Fix: `uv run ruff format examples/`

examples/todo/interface/server/bootstrap/todo_bootstrap.py
```diff
@@ -0,0 +1,17 @@
...
+from fastapi import FastAPI
+
+from examples.todo.infrastructure.di.todo_container import TodoContainer
+from examples.todo.interface.server.routers import todo_router
+
+
+def create_todo_container(todo_container: TodoContainer) -> None:
+    todo_container.wire(packages=["examples.todo.interface.server.routers"])
+
+def setup_todo_routes(app: FastAPI):
```
Mr-DooSun, line 13: **[CI] `ruff format` wants PEP 8 blank-line spacing** — two issues in this file:

1. Missing blank line after the module docstring (line 1 → line 2)
2. Only 1 blank line between top-level functions; `ruff format` enforces 2 blank lines

Compare to [`src/user/interface/server/bootstrap/user_bootstrap.py`](../../../../../src/user/interface/server/bootstrap/user_bootstrap.py) which has the correct spacing.

Fix: `uv run ruff format examples/` applies both automatically.

Mr-DooSun (commented):
Hi @Naman9245, one more follow-up — the `architecture` CI job is failing due to `ruff format` wanting to reformat 2 files (see 2 inline comments). Running `uv run ruff format examples/ tests/unit/todo/` locally and committing the result should make CI green. `ruff check` itself is already passing.

Mr-DooSun:
Hi @Naman9245, thanks again for this PR!

`main` has moved forward since this PR was opened — #116 was just merged, which touches several shared workflow docs and the layer dependency diagram. Could you please rebase or merge `main` into `feat/todo-example` so the PR re-runs CI against the current tree and any conflicts surface early?

No implementation changes to your todo example are expected — this is purely a sync request. Happy to help if anything conflicts.

Naman9245:
Hi @Mr-DooSun,

Done! I've rebased with main to sync the latest changes, manually removed the empty test files and the unused Optional imports, and ran both ruff format and ruff check --fix to resolve the CI spacing and import block issues.

Everything is completely green locally now. Thanks again for your patience and guidance!

Mr-DooSun (approved):
## LGTM 🎉

Hi @Naman9245,

Thank you for the patient and thoughtful work on this PR — five rounds of review with this much back-and-forth is a lot to navigate, especially as a first open-source contribution, and you handled every round with care, including the design discussion around `examples/` discovery and the SQLAlchemy 2.0 migration.

A few notes from the final pass:
- All 16 review threads are addressed and resolved
- CI is green across all 5 jobs (architecture / branch-name / lint / minimal-install / test)
- Branch is in sync with `main`; `ruff format --check` is clean
- The `Mapped[datetime]` typing on `created_at` / `updated_at` is actually more correct than what `UserModel` does today — nice catch on the upgrade

This sets a solid baseline for the four remaining example domains (url-shortener, blog, webhook-receiver, simple-chatbot). Thank you for setting the bar.

If this blueprint was helpful for your learning, a ⭐ on the repo would mean a lot — it really helps projects like this reach new contributors. Either way, hope to see more of your work out there 🙌

Issue #95, opened 2026-04-20
examples/blog: two domains with Protocol-based cross-domain DIP


Mr-DooSun:
Part of #85. Third of 5 seed examples.

## What to build

Two domains (`author`, `post`) where `post` depends on `author` **via
`AuthorRepositoryProtocol`, not via a direct import**. This is the
canonical pattern that keeps domains independently deployable.

## Acceptance criteria

- `examples/blog/author/` — standard domain layout, minimal fields (`id`, `display_name`)
- `examples/blog/post/` — standard domain layout (`id`, `author_id`, `title`, `body`)
- `examples/blog/post/domain/services/post_service.py` receives an `AuthorRepositoryProtocol` via DI, **not** `AuthorRepository`
- `examples/blog/post/infrastructure/di/post_container.py` wires the protocol to the concrete `AuthorRepository` from the `author` domain
- `GET /post/{id}` response includes the author's `display_name`, fetched via the protocol
- `examples/blog/tests/test_post_service.py` passes a mock implementing `AuthorRepositoryProtocol` (no `author` infrastructure needed)
- `examples/blog/README.md` — explains the protocol-based boundary in 1 paragraph

## Verification

1. Copy both `examples/blog/author/` and `examples/blog/post/` into `src/`
2. `make quickstart` — confirm startup auto-discovery picks both up
3. Create an author, then a post referencing that author
4. `GET /post/{id}` returns the author's `display_name`
5. `pytest examples/blog/tests/ -v` passes

## References

- Cross-domain skill: `$add-cross-domain` / `/add-cross-domain`
- ADR 007 (DI container + app separation): https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/docs/history/007-di-container-and-app-separation.md
- AGENTS.md cross-domain section: https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/AGENTS.md
- Tutorial: https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/docs/tutorial/first-domain.md
- Examples guide (landing in #92): https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/92

## Out of scope

Comments, markdown rendering, draft/publish states. One protocol, two domains, done.

PR #237, opened 2026-06-12
feat: add blog example with Protocol-based cross-domain DIP

Aydeing:
## Summary
- Two domains (`author`, `post`) demonstrating Protocol-based Dependency Inversion
- `post` depends on `author` via `AuthorRepositoryProtocol`, NOT direct import
- `GET /post/{id}` response includes `author_display_name` fetched via protocol
- Standard 10-file layout per domain following `examples/todo/` patterns
- Test with mock implementing `AuthorRepositoryProtocol` (no author infra needed)

Addresses Mr-DooSun/fastapi-agent-blueprint#95

## Test plan
- [x] `pytest examples/blog/tests/ -v` passes (2/2)
- [x] Verify no regressions in existing tests (941 passed, 11 pre-existing failures in `agents_shared/` unrelated to this PR)
- [ ] Copy to src/, `make quickstart`, create author + post, verify response includes author name


Mr-DooSun (changes requested):
Thanks for taking this one on, @Aydeing — the per-domain 10-file layout, the bootstrap wiring, and the protocol-injected service are all shaped correctly and mirror `examples/todo/` cleanly. Before this can land as the canonical cross-domain reference, though, there are a few blockers — most importantly, **the example doesn't actually run as written**. I did a careful pass plus a second independent review; consolidated below.

## 🔴 Blocking

### 1. `find_by_id` doesn't exist — every post path crashes at runtime
`examples/blog/post/domain/services/post_service.py:25`
```python
author = await self._author_repository.find_by_id(author_id)
```
`BaseRepository` / `BaseRepositoryProtocol` expose no `find_by_id`; the method is **`select_data_by_id`** (`src/_core/domain/protocols/repository_protocol.py`, `src/_core/infrastructure/persistence/rdb/base_repository.py`). Since `create_post`, `list_posts`, `get_post` **and** `update_post` all call `get_author_display_name`, copying this into `src/` and running the documented `make quickstart` → `GET /post/{id}` raises `AttributeError` (500) on every post path. That's the example's headline feature, and it fails issue #95's acceptance criterion ("`GET /post/{id}` returns the author's `display_name`").

One subtlety: `select_data_by_id` **raises 404** when the row is missing (it never returns `None`), so a straight rename would make the `else "Unknown"` branch (line 26) dead code and turn a missing author into an error on post reads. For the graceful fallback, use the no-raise batch read:
```python
async def get_author_display_name(self, author_id: int) -> str:
    authors = await self._author_repository.select_datas_by_ids([author_id])
    return authors[0].display_name if authors else "Unknown"
```
(Or keep `select_data_by_id` and drop the fallback if a dangling author should be a hard 404.)

## 🟠 High

### 2. The test never runs in CI — and it hides #1
`examples/blog/tests/test_post_service.py`
CI runs `pytest tests/` and `pyproject.toml` sets `testpaths = ["tests"]`, so `examples/blog/tests/` is **never collected**. The sibling example lives at `tests/unit/todo/domain/test_todo_service.py` — please mirror it at **`tests/unit/blog/domain/test_post_service.py`**.

Separately, the test uses a bare `AsyncMock()`, so `mock_author_repository.find_by_id` is auto-fabricated and the assertion passes green — which is exactly why #1 slipped through. After the fix, assert against the real method; `AsyncMock(spec=AuthorRepositoryProtocol)` would make a future method-name drift fail loudly.

(Heads-up: issue #95 itself specifies `examples/blog/tests/test_post_service.py` — that path is inconsistent with the contribution guide and CI. Following the guide is correct here; I'll align the issue text so it isn't misleading for the next contributor.)

### 3. It diverges from the cross-domain pattern it's meant to teach
`examples/blog/post/domain/protocols/author_repository_protocol.py`
The canonical procedure (`docs/ai/shared/skills/add-cross-domain.md`) and the only real precedent in the repo — `src/auth/domain/services/auth_service.py` importing `src.user.domain.protocols.user_repository_protocol` — both follow: **the provider owns the protocol; the consumer imports it.** This PR instead creates a second, byte-identical `AuthorRepositoryProtocol` inside `post/` (beside the one already in `author/domain/protocols/`), and the README frames that as the blueprint's pattern. A reader copying this learns a different pattern than `/add-cross-domain` teaches.

Suggested: delete `post/domain/protocols/author_repository_protocol.py` and import the author domain's protocol in `post_service.py`:
```python
from examples.blog.author.domain.protocols.author_repository_protocol import (
    AuthorRepositoryProtocol,
)
```
Also worth rewording: the consumer's protocol already imports the provider's `AuthorDTO`, so the README's "independently deployable — the only coupling is the DI container" overstates it. `post`'s domain layer does depend on `author`'s DTO type; the README should say so plainly.

## 🟡 Low / Polish

### 4. README verification won't run as written
`examples/blog/README.md` — the `curl` calls use `http://localhost:8000`, but quickstart serves on **8001** (`Makefile`, `run_server_local.py`). Also the endpoint tables show `/v1/author/{id}` while the routers correctly use `/{author_id}` and `/{post_id}` — please align the tables.

### 5. Request-schema bounds (consistency with `todo`)
`examples/blog/post/interface/server/schemas/post_schema.py` — `todo` bounds its free-text field (`description: max_length=1000`), but `body: str` here is unbounded, and `author_id: int` has no `Field(ge=1)`. Either add bounds for consistency, or — if an unbounded `Text` body is intentional — that's fine, just keep it deliberate.

## 💡 Optional (would make the example better)
`list_posts` does one author lookup per post (N+1) — since this is a reference, consider batching with `select_datas_by_ids` + a dict lookup. And because `post` already injects the author repository, a `_validate_create` hook calling `author_repository.exists_by_id(entity.author_id)` would demonstrate *why* you'd reach into another domain (not just to read a display name) — that's arguably the most instructive thing this example could show. Both are nice-to-haves, not blockers.

## ✅ Looks good
- Bootstrap naming (`bootstrap_{author,post}_domain`) matches the `_bootstrap_domains` auto-discovery contract exactly.
- Container shape, DTO base, and each domain owning its own repo protocol all match `examples/todo/`.
- No layer-boundary violations: no Domain→Infrastructure import, no Model leaking the repository, no Mapper/Entity class.
- Routes are unauthenticated — that's **consistent with `examples/todo/` and the examples profile** (auth is demonstrated in `src/auth`), so no change needed there.

Once #1–#3 are addressed (and ideally #4–#5), this'll be a solid reference. Thanks again for the contribution!

Aydeing:
Thanks for the thorough review @Mr-DooSun! Addressed all five and both optional items:

1. ✅ `select_datas_by_ids([id])` with `"Unknown"` fallback
2. ✅ Tests moved to `tests/unit/blog/domain/test_post_service.py`; `AsyncMock(spec=...)` on both protocols
3. ✅ Deleted duplicate protocol; consumer imports the author domain's; README rewritten to acknowledge the `AuthorDTO` coupling
4. ✅ Port 8001 + `/{author_id}` / `/{post_id}` paths
5. ✅ `author_id: Field(ge=1)`, `body: Field(max_length=10000)`
- 💡 `list_posts` now uses one batched `select_datas_by_ids` (no more N+1)
- 💡 `_validate_create` calls `ensure_existing_references` (existing helper in `src/_core/domain/validation.py`) — creating a post against a missing `author_id` now returns 422

Verified: `pytest tests/unit/blog/` → 6/6 pass; full unit suite → 463 pass, 0 regressions.

Mr-DooSun (dismissed):
## ✅ Round-2 review — all blockers verified resolved

Thanks for the thorough turnaround, @Aydeing. I re-reviewed against the actual framework contracts (not just "tests pass") and traced each fix back to the base classes. All 5 findings + both optional items check out.

**Verified fixed**

1. **Runtime crash** — `find_by_id` → `select_datas_by_ids` (a real method on `BaseRepositoryProtocol`, returns `list[DTO]`); the `"Unknown"` fallback is correct because the batch read doesn't 404 on a missing row. The headline `GET /v1/post/{post_id}` path now actually runs.
2. **CI collection + the masked bug** — test moved to `tests/unit/blog/domain/test_post_service.py`, mirroring `tests/unit/todo/domain/` (provably collected under pytest's `prepend` mode; unique basename, no clash). `AsyncMock(spec=...)` on both protocols now makes any method-name drift fail loudly — exactly the guard that was missing before.
3. **Cross-domain pattern** — duplicate protocol deleted; `post_service.py` imports the provider's `AuthorRepositoryProtocol`, and the DI wiring in `post_container.py` matches both `add-cross-domain.md` §3 and the live `auth_container.py` (`UserRepository` re-declaration). README now honestly states the `AuthorDTO` type coupling.
4. **README** — port 8001 + `/{author_id}` / `/{post_id}` paths.
5. **Schema bounds** — `author_id: Field(ge=1)`, `title`/`body` max-length bounded.

**Both optional items also landed:** `list_posts` is now a single batched `select_datas_by_ids` (no more N+1), and `_validate_create` uses the shared `ensure_existing_references` helper — actually nicer than a raw `exists_by_id` call since it reuses the canonical validation path. Confirmed the `create_data` → `_validate_create` → `insert_data` ordering: a missing author returns 422 and never inserts.

Layer boundaries, conversion patterns, container shape, and bootstrap naming are all clean. No security concerns beyond the unauthenticated example routes, which are the documented examples profile.

**Two optional nits (non-blocking — fine to land as-is or fold in):**
- `pyproject.toml` — the `"examples/**/tests/**/*.py" = ["S"]` ignore added earlier is now vestigial: the test moved out of `examples/`, so nothing matches that glob anymore (`tests/**` already covers `tests/unit/blog/`). Safe to drop.
- `examples/README.md` — the `blog/` catalog row is still 🟡; can flip to ✅ on merge (`todo/` is in the same boat, so no worries either way).

Nice work — approving. 🎉

Aydeing:
Pushed [61a491f](https://github.com/Aydeing/fastapi-agent-blueprint/commit/61a491f) — ran `ruff format` on the two flagged files (one-line signature/call collapse). Local check is now clean and the 6 blog tests still pass.

Heads-up on the other red check: the `Validate` (Governor Footer Lint) job failed with `Resource not accessible by integration` from `sticky-pull-request-comment` — that's the standard read-only `GITHUB_TOKEN` you get on `pull_request` events from a fork, so it isn't something I can fix from my side. Probably wants `pull_request_target` or a relaxed `permissions:` block on that workflow.

Mr-DooSun:
Thanks again for the solid work here, @Aydeing — and for the fast turnaround between review rounds. Everything's in good shape now: the cross-domain DIP lines up with the `auth → user` precedent, the test is collected and green, and the runtime path works as written.

I'm going ahead with the merge. The red `Validate` check is just the fork-PR `GITHUB_TOKEN` limitation you already spotted (it can't post the sticky comment) — not a code problem, safe to ignore. The two minor leftovers (the vestigial `pyproject` glob and the README catalog row) are tracked separately in #242, so they won't hold up your contribution.

If you've found the blueprint useful, a ⭐ on the repo would genuinely mean a lot — totally optional, though. Thanks for being a great contributor! 🙏

PR #239, opened 2026-06-14
feat(examples): add url-shortener example with Taskiq cleanup worker (#94)

### LINKED ISSUE #94: examples/url-shortener: CRUD + Taskiq cleanup worker
Part of #85. Second of 5 seed examples.

## What to build

A `link` domain plus a Taskiq worker task that deletes expired links.
Demonstrates how the **worker interface shares the same domain service**
as the HTTP router — write business logic once, expose it twice.

## Acceptance criteria

- Standard domain layout under `examples/url-shortener/` (same 10 files as the todo example)
- `LinkDTO(id, short_code, target_url, expires_at, created_at, updated_at)`
- Endpoints: `POST /link`, `GET /link/{short_code}`, `DELETE /link/{short_code}`
- `examples/url-shortener/interface/worker/tasks/cleanup_expired_links_task.py` — Taskiq task that calls `link_service.delete_expired(...)`
- `LinkService.delete_expired()` returns count of deleted rows
- `examples/url-shortener/tests/test_cleanup_task.py` — asserts expired rows are removed, non-expired rows remain
- `examples/url-shortener/README.md` — how to enqueue the cleanup task with `BROKER_TYPE=inmemory`

## Verification

1. Copy to `src/url_shortener/`
2. `make quickstart`
3. Create a link with `expires_at` in the past via `curl`
4. Invoke the cleanup task (enqueue via Python REPL or a tiny script)
5. Confirm the row is gone

## References

- Worker interface reference: https://github.com/Mr-DooSun/fastapi-agent-blueprint/tree/main/src/user/interface/worker
- Broker config: https://github.com/Mr-DooSun/fastapi-agent-blueprint/tree/main/src/_core/infrastructure/taskiq
- Tutorial: https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/docs/tutorial/first-domain.md
- Examples guide (landing in #92): https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/92

## Out of scope

Analytics, click tracking, custom slug validation, redirect responses. Focus is the worker pattern.

Ayushhh26:
## Related Issue
- Closes #94

## Change Summary
- Adds `examples/url-shortener/` — a CRUD `link` domain plus a Taskiq cleanup task that shares the same `LinkService` as the HTTP router (the worker pattern this issue targets).
- Mirrors the canonical `examples/todo/` layout (Domain / Infrastructure / Interface) and the production `src/user` worker pattern.
- Endpoints: `POST /v1/link`, `GET /v1/link/{short_code}`, `DELETE /v1/link/{short_code}`.
- Cleanup worker: `cleanup_expired_links_task` → `LinkService.delete_expired()` (idempotent bulk delete of expired rows).
- CI-discovered unit test under `tests/unit/url_shortener/` (logic in `examples/url-shortener/tests/`).
- README with curl walkthrough + InMemory-broker enqueue steps.

Out of scope (per issue #94): redirects, analytics/click tracking, custom slug validation.

## Type of Change
- [x] feat: New feature

## Checklist
- [x] Tests pass
- [x] Linting passes (`ruff check src/`)

## How to Test
- `cp -r examples/url-shortener src/url_shortener && rm -f ./quickstart.db && make quickstart`
- Exercise the curl commands in `examples/url-shortener/README.md`.
- `uv run pytest tests/unit/url_shortener/ -q`

---

[skip-governor-footer] Non-governor-changing PR — touches only `examples/` and `tests/`, no Tier A/B/C governor paths.


Mr-DooSun (commented):
Thanks for this, @Ayushhh26 — really nice work. It faithfully delivers what #94 asked for: one `LinkService` shared by an HTTP router and a Taskiq worker, with the cleanup task reusing domain logic instead of duplicating it. I reviewed it against the project's architecture + security checklists under the **examples profile**, and it's in good shape — **no blocking issues**. One maintainability suggestion and a few optional notes below.

## ✅ What's solid

- **The worker pattern is exactly right.** `cleanup_expired_links_task` is a parameterless `@broker.task` delegating to `LinkService.delete_expired()` — the same shape as the production `src/_apps/worker/tasks/audit_cleanup_task.py`, so there's no payload to validate and the bulk delete is idempotent. 👍
- **Clean layering.** No `infrastructure` / `sqlalchemy` / `dependency_injector` imports in `domain/`; the schema import into the service is the permitted Request-type path.
- **Conversion + DI are correct.** 3-TypeVar `BaseService[CreateLinkRequest, UpdateLinkRequest, LinkDTO]`, repo returns `LinkDTO.model_validate(..., from_attributes=True)` and keeps `LinkModel` inside the repository, `DeclarativeContainer` with `Singleton` repo + `Factory` service.
- **Self-contained on copy.** The relative imports + `wire(packages=["src.url_shortener..."])` targets mean the package works standalone once copied into `src/url_shortener/`, without depending on `examples/` staying around. (Please keep these — see the "don't change" note below.)
- **Auth-free POST/DELETE is fine here** — the examples profile explicitly allows it; this example teaches the worker pattern, not auth.

## 🔧 Suggestion (not a merge blocker): rename the directory to an importable name

The folder name `url-shortener` (hyphen) isn't a valid Python package name, and that single fact is what forces all the test scaffolding:

- the `sys.modules` package-manufacturing block in `examples/url-shortener/tests/test_cleanup_task.py:16-39`, and
- the second `importlib.util.spec_from_file_location` loader in `tests/unit/url_shortener/test_cleanup_task.py`.

The existing `examples/todo/` example needs none of that — its test imports the package directly.

Renaming to an underscore fixes the root cause:

```bash
git mv examples/url-shortener examples/url_shortener
```

Then the test collapses to a single direct-import file, mirroring `tests/unit/todo/domain/test_todo_service.py`:

```python
# tests/unit/url_shortener/test_cleanup_task.py  (after rename — no importlib / sys.modules)
from examples.url_shortener.domain.dtos.link_dto import LinkDTO
from examples.url_shortener.domain.services.link_service import LinkService
# ... build an in-memory repo, assert delete_expired() drops only expired rows
```

and `examples/url-shortener/tests/` can be deleted (todo has no in-example test dir; the contributing guide only asks for `tests/unit/{name}/`). Bonus: the README install step becomes the friction-free same-name copy `cp -r examples/url_shortener src/url_shortener`, with no hyphen→underscore explanation needed.

**Please _don't_ change** the relative imports or the `src.url_shortener` wire targets — those are the good part and should stay. Only the directory name needs to change.

## Optional minor notes (take or leave)

- **tz normalization is one-sided.** `examples/url-shortener/infrastructure/repositories/link_repository.py` adds `_to_naive_utc` and applies it on the delete path, but the create path stores `CreateLinkRequest.expires_at` as-is. With the documented naive inputs on SQLite this never bites, but a tz-aware `expiresAt` could store inconsistently (and raise against a Postgres `timestamp without time zone` column). A field validator normalizing `expires_at` to naive UTC — or a one-line README note that inputs are treated as naive UTC — would close the asymmetry.
- **Optional `schedule=` label.** Adding `schedule=[{"cron": "..."}]` to the task (like `audit_cleanup_task`) would also demonstrate the scheduled-cleanup path. The current manual-`.kiq()` walkthrough is a perfectly valid teaching choice, so this is purely additive.
- **`target_url` validation** is only worth tightening (`HttpUrl`) if redirects are ever added — which #94 explicitly puts out of scope, so no action needed now.

Happy to merge once the rename is in — or if you'd rather keep the hyphenated layout, let's talk it through here first. Thanks again for the clean, well-documented contribution! 🙏

Ayushhh26:
Thanks for the feedback! Just pushed an update to address everything.

I renamed the directory to url_shortener, which let me completely scrap that sys.modules test hack and use clean, direct imports. I also updated the README copy instructions and added the quick note about naive UTC inputs.

As requested, I left the relative imports and wire targets exactly as they were. Let me know if this is good to go!

Mr-DooSun:
Thanks for the quick turnaround, @Ayushhh26 — the rename is exactly right. Both test shims are gone, the test now imports the package directly like `examples/todo`, the copy step is the friction-free same-name copy, and you kept the relative imports + `src.url_shortener` wire targets intact as asked. The naive-UTC README note closes the tz point too. Round-1 feedback is fully addressed. ✅

A deeper cross-review surfaced **two new things** (neither was in my first pass) — both minor, neither blocks merge:

### 1. The InMemory worker walkthrough doesn't actually run as written — and this one's on us, not you 🙏

The README's "Enqueue Cleanup With InMemory Broker" section says to start a worker in a second terminal with `BROKER_TYPE=inmemory ... run_worker_local.py`, then `.kiq()` from a REPL. I ran it, and the worker **crash-loops** on:

```
RuntimeError: Inmemory brokers cannot listen.
  File ".../taskiq/brokers/inmemory_broker.py", line 180, in listen
```

Taskiq's `InMemoryBroker` is process-local — `listen()` is unsupported and `kick()` runs the task **inline in the calling process**, so a separate worker never receives a REPL-enqueued task (and a bare REPL also doesn't wire the DI container). You faithfully followed the pattern our own `.claude/rules/commands.md` documents as the worker "default", so this is **pre-existing doc debt on our side** — now tracked in #241.

For the example itself, two ways to make the walkthrough real (your pick):
- **(A) zero-config, single process:** drop the second-terminal worker; have the snippet bootstrap + wire the worker container, `.kiq()`, then `await broker.wait_all()` — all in one script.
- **(B) real separate worker:** demo it with `BROKER_TYPE=rabbitmq` (matches the genuine queued-worker experience).

Happy to take this README fix as a follow-up ourselves if you'd rather wrap up here — your call.

### 2. Bound the `short_code` path param (trivial)

`CreateLinkRequest.short_code` caps at 64 and the column is `String(64)`, but `GET`/`DELETE /v1/link/{short_code}` accept an unbounded `str`. One-liner on both handlers:

```python
from typing import Annotated
from fastapi import Path

short_code: Annotated[str, Path(min_length=1, max_length=64)]
```

The code itself is clean and the unit test passes — thanks again for a well-structured contribution. 🙏

Mr-DooSun:
Hi @Ayushhh26 — Round-1 is fully wrapped (the `url_shortener` rename killed the import shims, the test imports the package directly, CI's green). 🙏

We just merged #243, which fixes the project-wide InMemory-worker doc debt (#241) and adds a `make worker` fast-fail guard — so the two items left here are small. I've written them as a precise checklist in case you (or a coding agent like Claude Code / Codex) want to knock them out quickly. The repo's `AGENTS.md` + the "examples profile" rules apply.

### ☐ 1. Fix the README worker walkthrough (it currently crash-loops)
- **File:** `examples/url_shortener/README.md` → the **"Enqueue Cleanup With InMemory Broker"** section.
- **Problem:** it tells the reader to start a standalone worker on `BROKER_TYPE=inmemory` (`run_worker_local.py` in a 2nd terminal). InMemory can't `listen()`, so the worker crash-loops (`RuntimeError: Inmemory brokers cannot listen.`). InMemory runs tasks **inline in the producer process** — there is no separate worker to receive a `.kiq()`.
- **Fix — Option A (zero-config, single process; recommended):** drop the 2nd-terminal worker; have the snippet bootstrap the wired broker, kick the task, and await it inline:
  ```python
  result = await cleanup_expired_links_task.kiq()
  await result.wait_result()
  ```
  This is the exact inline pattern the sibling example already uses — see `tests/unit/webhook_receiver/test_webhook_event.py` (`kiq()` → `wait_result()`). Keep it consistent with the now-fixed `docs/canonical-demo.md` Step 5 / `.claude/rules/commands.md` Broker section from #243.
- **Fix — Option B (optional):** instead, demo a real cross-process worker with `BROKER_TYPE=rabbitmq`.
- **Verify:** following the README top-to-bottom runs to completion (no crash-loop) and the expired link is deleted.

### ☐ 2. Bound the `short_code` path param (trivial)
- **File:** the `GET` / `DELETE /v1/link/{short_code}` handlers in `examples/url_shortener/interface/server/routers/link_router.py`.
- **Fix:**
  ```python
  from typing import Annotated
  from fastapi import Path

  short_code: Annotated[str, Path(min_length=1, max_length=64)]
  ```
  (matches the `CreateLinkRequest.short_code` cap and the `String(64)` column.)

### ☐ 3. Sync with `main` (#243 just landed)
- `git merge main` (or rebase). Expect a tiny conflict in `examples/README.md`: #243 flipped the `todo/` and `blog/` catalog rows to ✅ on the lines right around your `url_shortener/` row. Keep all three, and flip the `url_shortener/` row to ✅ while you're there.
- **Verify:** `make check` stays green.

No pressure on who does this — happy for you to take it, or if you'd rather call it done here, we're glad to land these and merge with credit to you. Just let us know 🙏

Ayushhh26:
Thanks for the updated checklist! I've gone ahead and wrapped up all three items:

Merged main and resolved the README.md conflict.

Updated the worker walkthrough to use the inline Option A.

Added the Path bounds to the short_code parameter.

make check is completely green locally. Let me know if this is good to go!

Mr-DooSun:
Thanks for the fast turnaround, @Ayushhh26 🙏 Two of the three are landed cleanly:

- ✅ **#2 path bound** — `GET` / `DELETE /v1/link/{short_code}` now use `Annotated[str, Path(min_length=1, max_length=64)]`.
- ✅ **#3 main sync + catalog** — merge is clean and the `examples/README.md` catalog row is flipped, with `todo` / `blog` / `url_shortener` all showing ✅.

I ran **#1** top-to-bottom before merging, and the README worker walkthrough still doesn't complete as written. Running the exact snippet from the *"Enqueue Cleanup With InMemory Broker"* section (inmemory + sqlite — the quickstart defaults):

```
AttributeError: 'Provide' object has no attribute 'delete_expired'
```

The task errors, so the expired row is **not** deleted. `wait_result()` doesn't re-raise, so the script itself exits 0 — you just see taskiq's traceback and the final `curl` still returns the link.

**Why:** with `BROKER_TYPE=inmemory`, `.kiq()` runs the task **inline in the calling process**, so *that* process needs the container wired. The bare `uv run python` snippet only imports the task and calls `.kiq()` — it never instantiates/wires `UrlShortenerContainer`, so `@inject` hands the raw `Provide` marker to `link_service`. Adding a one-line `container.wire(...)` isn't enough either: it then hits `Dependency "UrlShortenerContainer.core_container.database" is not defined`, because the DB provider is only set up by the full worker bootstrap.

The working reference for this exact shape is already in the repo and green — [`tests/unit/webhook_receiver/test_webhook_event.py`](https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/tests/unit/webhook_receiver/test_webhook_event.py): it instantiates the container, sets up the DB, `container.wire(modules=[...])`, **then** `kiq()` → `wait_result()`. A single-process README snippet needs that same bootstrap, not just the two-liner.

So the two honest options are:
- **(A) single-process:** the snippet bootstraps + wires the container (mirroring the webhook test) before `.kiq()`.
- **(B) real cross-process worker:** `BROKER_TYPE=rabbitmq` + `run_worker_local.py` (now that #243's guard fast-fails inmemory workers), and `.kiq()` from a small producer — there the producer only needs the broker, and the wired worker process runs the task.

Since #2/#3 are clean and this last one is really a bootstrap quirk on our side, we're glad to **land the README fix ourselves and merge with credit to you** — or if you'd like one more quick pass at it, go for it. Either way, thanks for a genuinely well-structured contribution 🙏

Ayushhh26:
I think it would be much better if you make this final README fix. Since you know exactly how you want the code to look in your documentation, it will be much faster and avoid another round of review.

Please go ahead and apply it and merge! Thank you so much for all your help and patience with me on this PR.

Mr-DooSun:
Thanks so much for this, @Ayushhh26! 🙏

I've pushed the final README fix (`27045a1`) and I'll merge it now — verified it runs top-to-bottom on the quickstart defaults (`cleanup task deleted 1 expired link(s)`, and the permanent link stays untouched). That last piece was just a DI-bootstrap quirk on our side, nothing to do with your code.

This was a genuinely well-structured contribution — you mirrored the canonical example layout cleanly and turned around the path-bound and main-sync fixes fast. Really appreciate you taking an interest in the project and putting in the work.

Hope you'll give the blueprint a spin in your own projects sometime, and the door's always open if you'd like to contribute again. Thanks again! 🎉

Ayushhh26:
Thanks so much for all the guidance and help with this feature. It was a great experience contributing, and I'll definitely be keeping an eye on the project.

PR #249, opened 2026-06-20
feat: add simple-chatbot example (#97)

vraj2131:
## Related Issue
- Fixes #97

## Change Summary
- Scaffolded a stateless simple chatbot example under `examples/simple-chatbot/` leveraging a PydanticAI Agent with a structured output type `ChatReply` and an inline system prompt.
- Persisted conversation history to the database using `ChatMessageDTO`.
- Calculated and surfaced total token usage (`tokens_used = input_tokens + output_tokens`) from agent run results for educational visibility.
- Configured a dynamic `sys.meta_path` redirect finder in `examples/simple-chatbot/tests/conftest.py` so unit tests can be run standalone within the example folder.
- Added global unit tests under `tests/unit/simple_chatbot/`.
- Updated status for the available examples table in `examples/README.md`.

## Type of Change
- [x] feat: New feature

## Checklist
- [x] Architecture rules followed (no Domain -> Infrastructure imports)
- [x] Tests pass
- [x] Linting passes (`ruff check src/`)

## How to Test
1. Run the standalone unit tests:
   ```bash
   pytest examples/simple-chatbot/tests/ -v
   ```
2. Copy the example to `src/` to test manually:
   ```bash
   cp -r examples/simple-chatbot src/simple_chatbot
   pytest tests/unit/simple_chatbot/ -v
   ```
3. Boot the quickstart server:
   ```bash
   make quickstart
   ```
   Submit a query:
   ```bash
   curl -X POST http://127.0.0.1:8001/v1/chat -H "Content-Type: application/json" -d '{"prompt": "Hello"}'
   ```
   Retrieve history by ID:
   ```bash
   curl http://127.0.0.1:8001/v1/chat/1
   ```

[skip-governor-footer]

PR #253, opened 2026-06-26
docs(examples): distinguish DB-only examples from real-LLM-calling examples

### LINKED ISSUE #252: docs(examples): distinguish DB-only examples from real-LLM-calling examples
## Context

`examples/simple_chatbot/` (#249) is the first example that calls a **real external LLM**, which introduces a cost/abuse surface the pure-DB examples (`todo`, `url_shortener`, `webhook_receiver`) do not have. All examples are intentionally unauthenticated reference code, but a reader copying an LLM-calling example into production inherits a meaningfully different risk profile.

## Proposal

Add a short note to `examples/README.md` (and/or a shared callout) distinguishing:

- **DB-only examples** — safe to copy as-is for evaluation
- **LLM-calling examples** — require auth + per-user rate/budget caps + guardrails before exposing externally

The `simple_chatbot` README already carries a per-example "Production Considerations" note; this lifts the distinction up to the examples index so it is visible *before* a reader picks an example.

## Why this is a follow-up

It is a project-level documentation decision, not the contributor's job, and not a defect in #249.

ZachDreamZ:
The examples README didn't distinguish between examples safe to copy as-is and those that call external LLMs with a cost/abuse surface.

- Added a production-surface callout before the Available examples table
- Added a Surface column to the table (DB-only / LLM-calling)

Closes #252

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -85,17 +85,24 @@ Architectural rules in [`AGENTS.md`](../AGENTS.md) apply to examples
 just like production domains — no `Domain → Infrastructure` imports,
 no Model objects leaving the Repository, no Mapper classes.
 
+> ⚠️ **Production surface**
+>
+> DB-only examples are safe to copy as-is for local evaluation.
+> LLM-calling examples call a real external provider and introduce a
+> cost/abuse surface. Before exposing one externally, add auth,
+> per-user rate/budget caps, and runtime guardrails.
+
 ## Available examples
 
 Populated incrementally as contributors land the good-first-issues:
 
-| Example | Pattern it teaches | Status |
-|---|---|---|
-| `todo/` | Minimal CRUD domain (zero infra) | ✅ [#112](https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/112) |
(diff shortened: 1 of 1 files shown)


Mr-DooSun:
Reviewed against the shared architecture/security rules and governor-paths.md.

Verdict: merge-ready. Docs-only change to examples/README.md — no findings, no governed/Tier-1 path touched, sync not required. The "LLM-calling" warning ahead of any landed LLM example is not a doc-vs-reality issue: the table already lists simple-chatbot as a tracked (unlanded) example.

Coordination note: this PR and #249 both edit the simple-chatbot row of the examples table from the same base, so a textual conflict is expected for whichever lands second. This PR is mergeable now and #249 is currently blocked, so landing this first is fine. When #249 is later rebased, keep the new Surface column, rename the row to `simple_chatbot/`, keep `LLM-calling`, and flip the status to ✅ [#249].

Mr-DooSun:
Thank you so much for picking this up, @ZachDreamZ! #252 was actually one I'd left for myself to get to later, so I honestly didn't expect it to be handled this quickly. I really appreciate you taking an interest in the project and putting in the work here.

Merging this now — thanks again! 🙏

PR #267, opened 2026-07-02
test: add HTTP middleware contract tests (CORS / X-Request-ID / ordering)

### LINKED ISSUE #2: Expand test coverage
### Problem

Current test coverage is limited, making it difficult to catch regressions and validate behavior across all domains and layers.

### Proposed Solution

Increase unit and integration test coverage across all layers (Router → Service → Repository) with focus on:
- Edge cases and error handling paths
- Cross-domain interaction scenarios
- Middleware behavior (exception handling, CORS)
- Repository CRUD operations with test database

### Area

Other

### Alternatives Considered

_No response_

Diyaaa-12:
test: add HTTP middleware contract tests (CORS / X-Request-ID / ordering)

Covers the app-level middleware wired in src/_apps/server/bootstrap.py:
- CORS preflight and simple-request headers
- X-Request-ID generation (absent) and echo (valid UUID4 provided)
- Middleware registration order: CorrelationId -> RequestLog -> CORS -> TrustedHost

## Related Issue
- Closes #2

## Change Summary
- Added `tests/e2e/test_middleware.py` covering CORS headers (preflight and simple request), X-Request-ID generation/echo via asgi-correlation-id, and the middleware registration order contract in `bootstrap.py`.

## Type of Change
- [ ] feat: New feature
- [ ] fix: Bug fix
- [ ] refactor: Code restructuring
- [ ] docs: Documentation
- [ ] chore: Build/tooling
- [x] test: Tests
- [ ] ci: CI/CD
- [ ] perf: Performance
- [ ] style: Code style

## Checklist
- [x] Tests pass
- [x] Linting passes (`ruff check src/`)

## How to Test
- `uv run pytest tests/e2e/test_middleware.py -v`
- `uv run ruff check tests/e2e/test_middleware.py`

---

## Notes
- `access-control-allow-origin` differs by request type: preflight reflects the request `Origin`, while simple requests return `*`. Both are asserted as observed.
- Followed the existing async test convention in `tests/e2e/` (explicit `@pytest.mark.asyncio` alongside `asyncio_mode = "auto"`).
- Ran the full suite locally (Windows) as a sanity check — all failures there are pre-existing, Windows-platform-specific (missing `sh`/`bash`, symlink privilege requirements, missing optional deps like `nicegui`/`boto3`/`aioboto3`), and unrelated to this change.


Diyaaa-12:
All checks green, ready for review whenever you get a chance! Let me know if anything needs adjusting.

tests/e2e/test_middleware.py
```diff
@@ -0,0 +1,79 @@
...
+    async with _client() as client:
+        response = await client.options(
+            "/health",
+            headers={
+                "Origin": "http://example.com",
+                "Access-Control-Request-Method": "GET",
+            },
+        )
+    assert response.status_code == 200
+    assert response.headers.get("access-control-allow-origin") == "http://example.com"
```
Mr-DooSun, line 35: **[NOTE] Preflight contract — pin `allow-methods`, and mind that this origin assertion is config-coupled.**

Two optional hardenings for the one test whose job is the preflight response:

1. **`allow-methods` is the other half of a preflight contract** but is left unasserted. The response returns `access-control-allow-methods: DELETE, GET, HEAD, OPTIONS, PATCH, POST, PUT` (the expansion of `allow_methods=["*"]` in `bootstrap.py`). A regression narrowing it wouldn't be caught here:
   ```python
   assert "GET" in response.headers.get("access-control-allow-methods", "")
   ```
2. **Config coupling:** `== "http://example.com"` only holds because `settings.allow_origins` defaults to `["*"]` — with `allow_credentials=True` that makes Starlette reflect the requested Origin. Settings load from `.env` and no conftest pins `ALLOW_ORIGINS`, so a dev/CI that sets a non-wildcard `ALLOW_ORIGINS` would flip this to a failure (the origin would no longer be allowed). Consider pinning it in-test so the contract is environment-independent:
   ```python
   monkeypatch.setattr(settings, "allow_origins", ["*"])
   ```

(See the comment on the simple-request test for the missing `allow-credentials` assertion, which applies here too.)

tests/e2e/test_middleware.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@pytest.mark.asyncio
+async def test_cors_headers_on_simple_request():
+    async with _client() as client:
+        response = await client.get("/health", headers={"Origin": "http://example.com"})
+    assert response.status_code == 200
+    assert response.headers.get("access-control-allow-origin") == "*"
```
Mr-DooSun, line 46: **[NOTE] Simple-request CORS — env-coupled `*`, a missing credentials assertion, and an untested reflection branch.**

1. **Config coupling:** `== "*"` depends on the `allow_origins=["*"]` default (same `.env` / no-conftest-pin coupling as the preflight test). A non-wildcard `ALLOW_ORIGINS` makes Starlette reflect the origin instead of returning `*`, breaking this assertion.
2. **Most useful add — pin the credentials flag:** neither CORS test asserts `access-control-allow-credentials`, yet `allow_credentials=True` (`bootstrap.py:79`) is the security-sensitive property that makes a wildcard origin dangerous and is the reason the strict-env validator (`config.py`, #227) exists. A regression dropping it stays green here. This assertion is stable regardless of `ALLOW_ORIGINS`:
   ```python
   assert response.headers.get("access-control-allow-credentials") == "true"
   ```
3. **Untested credentialed-reflection path:** with `allow_credentials=True` + wildcard, Starlette flips ACAO from `*` to the reflected `Origin` the moment a `Cookie` is present (`CORSMiddleware.send`: `allow_all_origins and has_cookie`). That's the actual risk the strict-env validator guards, and it's never exercised. Optional extra case:
   ```python
   async def test_cors_reflects_origin_when_cookie_present():
       async with _client() as client:
           response = await client.get(
               "/health",
               headers={"Origin": "http://example.com", "Cookie": "session=x"},
           )
       assert response.status_code == 200
       assert response.headers.get("access-control-allow-origin") == "http://example.com"
       assert response.headers.get("access-control-allow-credentials") == "true"
   ```

tests/e2e/test_middleware.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# --- X-Request-ID (asgi-correlation-id) ------------------------------------
+@pytest.mark.asyncio
+async def test_request_id_generated_when_absent():
+    async with _client() as client:
+        response = await client.get("/health")
+    assert response.headers.get("X-Request-ID")  # a UUID4 hex is generated
```
Mr-DooSun: **[LOW] This assertion doesn't check what its comment promises.**

`assert response.headers.get("X-Request-ID")` passes for *any* non-empty string, but the comment says "a UUID4 hex is generated". `CorrelationIdMiddleware`'s generator is `lambda: uuid4().hex`, so a regression to a static token, a counter, or a truncated id would silently keep this test green. The sibling echo test already pins an exact UUID4 — so the *generation* path is the one left unguarded, which is exactly what this test is for. Strengthen it to the contract the comment describes:

```python
# before
assert response.headers.get("X-Request-ID")  # a UUID4 hex is generated

# after  (add `import uuid` at module top)
request_id = response.headers.get("X-Request-ID")
assert request_id is not None
assert uuid.UUID(request_id).version == 4  # matches asgi-correlation-id's uuid4().hex
```

`uuid.UUID()` accepts the 32-char hex form, so no reformatting is needed (verified against the live header value).

tests/e2e/test_middleware.py
```diff
@@ -0,0 +1,79 @@
...
+    from fastapi.middleware.cors import CORSMiddleware
+    from fastapi.middleware.trustedhost import TrustedHostMiddleware
+
+    from src._core.infrastructure.logging.request_log_middleware import (
+        RequestLogMiddleware,
+    )
+
+    assert [m.cls for m in app.user_middleware] == [
+        CorrelationIdMiddleware,
+        RequestLogMiddleware,
```
Mr-DooSun, line 93: **[NOTE] `RequestLogMiddleware` is in the order assertion but has no behavioral coverage.**

The docstring names the "logging order the rest of the app depends on", and the whole reason `CorrelationIdMiddleware` is added *after* `RequestLogMiddleware` (`bootstrap.py:70-84`) is so the emitted log line carries `X-Request-ID`. That "the order produces the intended effect" claim is asserted only structurally here.

Optional (lower priority — the order + `X-Request-ID` tests already give indirect protection): since `tests/conftest.py` already resets structlog for deterministic `capture_logs`, one test could GET `/health` and assert the emitted `http_request` log event's bound `request_id` equals `response.headers["X-Request-ID"]`, binding the ordering rationale to observed behavior.

tests/e2e/test_middleware.py
```diff
@@ -0,0 +1,79 @@
...
+        CORSMiddleware,
+        TrustedHostMiddleware,
```
Mr-DooSun, line 95: **[NOTE] `TrustedHostMiddleware` is checked for presence only — its reject path is never exercised.**

Every request uses `base_url="http://localhost"`, which always passes the host check, so the middleware's actual purpose (rejecting hosts outside `allowed_hosts=["localhost","127.0.0.1"]`) is untested. A regression widening `allowed_hosts` to `["*"]` would pass every test in this file. One request closes the gap (verified: a bogus Host returns `400 Invalid host header`):

```python
async def test_rejects_untrusted_host():
    async with _client() as client:
        response = await client.get("/health", headers={"Host": "evil.example.com"})
    assert response.status_code == 400  # TrustedHostMiddleware rejects hosts outside allowed_hosts
```

Mr-DooSun (commented):
Nice, high-value regression coverage — the middleware **registration order** in particular is subtle (Starlette applies the last-added middleware as the outermost, and the `CorrelationId` → `RequestLog` ordering rationale is documented in `bootstrap.py`), and pinning it is exactly the kind of contract that breaks silently. The file also faithfully mirrors the existing `tests/e2e/test_docs_routes.py` convention (`_client()` helper + explicit `@pytest.mark.asyncio`), and all 5 tests pass locally.

This is **merge-ready** — every inline comment below is an optional `NOTE`/`LOW` hardening, none block merge. The common thread: the order test proves each middleware is *registered*, and the behavioral tests cover CORS-origin + `X-Request-ID`, but a few of the *behaviors* the file frames itself as guarding are left half-asserted (the credentials flag, the TrustedHost reject path, `allow-methods`, and the request-log effect).

Two things I looked at and would **keep exactly as-is**:
- The function-local middleware imports in `test_middleware_registration_order` are fine — scoping test-only deps to the single test that uses them is a reasonable idiom, no change needed.
- The exact-list `==` in the order assertion is a **strength**, not a fragility: it also catches a *dropped/removed* middleware, and a count mismatch on a test with this name is trivially diagnosable. Please **don't** loosen it to a relative / `index()` comparison — that would weaken the contract.

tests/e2e/test_middleware.py
Diyaaa-12: Added both — access-control-allow-methods is now asserted, and allow_origins is pinned via monkeypatch so this is environment-independent. Also added access-control-allow-credentials per your note on the sibling test.

tests/e2e/test_middleware.py
Diyaaa-12: Added the credentials assertion and the cookie-reflection test as a new case (test_cors_reflects_origin_when_cookie_present). allow_origins is now pinned here too.

tests/e2e/test_middleware.py
Diyaaa-12: Tried this — ran into an interaction between capture_logs() and the stdlib-bridged logger (structlog.stdlib.get_logger) where merge_contextvars didn't land in the captured event dict even with an explicit processor chain, despite the real console output showing request_id correctly. Since you'd flagged this as optional/lower-priority given the order + request-id tests already give indirect coverage, I didn't want to sink more time chasing the stdlib bridge interaction. Happy to revisit if you'd like it in scope for this PR.

tests/e2e/test_middleware.py
Diyaaa-12: Added test_rejects_untrusted_host covering the reject path with a bogus Host header.

tests/e2e/test_middleware.py
Diyaaa-12: Good catch — updated to assert uuid.UUID(request_id).version == 4 instead of just truthiness.

Mr-DooSun:
Merged — thank you, @Diyaaa-12! 🎉

This is a genuinely nice contribution and a welcome addition to the suite:

- Real contract coverage for the middleware stack — CORS (preflight / simple / cookie-reflect), X-Request-ID generation + echo, and the registration-order guard. This is exactly the kind of behaviour that breaks silently and is painful to catch later.
- The registration-order assertion is spot on: `app.user_middleware` is the reverse of the `add_middleware` call order, and your expected list matches `bootstrap.py` and the execution-order comment there.
- Clear module docstring, and thanks for documenting the preflight-vs-simple `Access-Control-Allow-Origin` difference and the async-test convention in the PR notes.

Verified locally before merge: 7/7 pass, and the full e2e suite stays green.

One tiny, entirely optional follow-up for whenever you're next in here (not a blocker — already merged): the `monkeypatch.setattr(settings, "allow_origins", ["*"])` in the three CORS tests is effectively a no-op. The app (and its `CORSMiddleware`) is built once at import in `app.py` (`app = create_app()`), so `CORSMiddleware` captures `allow_origins` at construction and never re-reads `settings` afterwards — and `["*"]` is already the default. The assertions still pass and still guard the real behaviour; the patch line (plus the now-unused `monkeypatch` parameter) can simply be dropped for clarity. If you ever want to pin a non-default `ALLOW_ORIGINS`, it has to be set in the environment before the app module is imported, since a post-import settings mutation can't reach the already-built middleware.

Thanks again for closing #2 — great first middleware-contract test set. 🙏

Issue #199, opened 2026-05-27
feat(auth): server-route RBAC for /v1/* endpoints


Mr-DooSun:
## Background

#194 landed page-level RBAC for NiceGUI admin pages (`User.permissions` + `require_auth(page_key=...)`). However, FastAPI API routes under `/v1/*` have no role or permission enforcement today — any authenticated user (role `user` or `admin`) can call any endpoint.

## Goal

Add per-endpoint role/permission gating to FastAPI routes so that sensitive operations (e.g. user management CUD, admin-only reads) require the caller to hold the appropriate role.

## Proposed Approach

- Introduce a `require_role(...)` dependency (or decorator pattern) that inspects the JWT-authenticated user's `role` field and raises `403 Forbidden` if the role requirement is not met
- Wire it into existing routes as needed — at minimum: `POST/PUT/DELETE /v1/user/...` restricted to `admin` role
- Consider whether `User.permissions` page keys are also relevant here or if role-only gating is sufficient for API routes
- Ensure the existing `test_auth_router.py` 401 tests still pass and add new 403 tests for role-gated routes

## Scope

- In scope: `/v1/*` FastAPI route-level role enforcement
- Out of scope: NiceGUI admin page permissions (already done in #194)

## Related

- Closes the "Server-route RBAC" item in `.claude/rules/project-status.md` § Not Yet Implemented
- Follows up on #194 page-level admin RBAC

PR #293, opened 2026-07-15
test: add Locust performance-test harness (#3)

### LINKED ISSUE #3: Add Locust performance testing
### Problem

There is no performance testing infrastructure to measure API response times, throughput, and behavior under load before deploying to production.

### Proposed Solution

Integrate Locust as the performance testing framework with:
- Pre-configured test scenarios for common API patterns (CRUD operations, concurrent requests)
- Makefile target for easy execution (`make perf-test`)
- Baseline performance metrics documentation

### Area

DevOps / CI/CD

### Alternatives Considered

_No response_

Ayushhh26:
## Related Issue
- Closes #3

## Change Summary
- `tests/perf/locustfile.py` — Locust harness with three scenarios: an always-on customer auth flow (`/v1/auth/register` → `/v1/auth/me` → `/v1/auth/refresh`, `/v1/auth/logout` on stop), always-on concurrent health reads (`/health` and `/health/db` as separate stat rows), and an env-gated admin CRUD flow on `/v1/user` (paginated list + create → read → update → delete with grouped `/v1/user/[id]` stats and self-cleanup)
- `make perf-test` — headless run against a local server; `PERF_HOST` / `PERF_USERS` / `PERF_SPAWN_RATE` / `PERF_RUN_TIME` overridable; exits non-zero if any request failed
- `docs/operations/performance-locust.md` — how to run, how to read the output, the one-time admin provisioning flow, a measured illustrative local baseline, and limitations (quickstart-local, illustrative only, no CI — per the scoping comment)
- Index updates: `docs/README.md` Operations table row + roadmap checkbox in `docs/reference.md`
- No `src/` changes, no new dependencies (locust was already in the dev group), no CI wiring

## Type of Change
- [x] test: Tests
- [x] docs: Documentation

## Checklist
- [x] Architecture rules followed (no Domain -> Infrastructure imports) — no `src/` changes
- [x] Tests pass — 1371 passed; one pre-existing failure unrelated to this PR (see Notes)
- [x] Linting passes (`ruff check src/`) — plus `ruff check tests/perf/` and all pre-commit hooks on the changed files

## How to Test
```bash
make quickstart      # terminal 1
make perf-test       # terminal 2 — 30s headless run, exits 0 with a stats summary
```
- Default run needs zero setup (customer + health scenarios). Measured here: 327 requests / 0 failures, 10.9 RPS, p50 3 ms / p95 25 ms / p99 220 ms — recorded in the docs as an illustrative baseline.
- Admin scenario: provision a real admin once via the dashboard (documented three-step flow), then `LOCUST_ADMIN_USERNAME=… LOCUST_ADMIN_PASSWORD=… make perf-test`.
- Negative probe: `LOCUST_ADMIN_USERNAME=admin LOCUST_ADMIN_PASSWORD=admin make perf-test` — the bootstrap admin is rejected by the token API, the login failure shows loudly in Locust's error report, and `make` exits non-zero.

## Scenario shape — as requested, pinging for your input (@Mr-DooSun)
- The requested "auth + CRUD flow on `/v1/user`" hit a wall: all `/v1/user` routes are admin-realm-gated (#199/#218), and the bootstrap admin is setup-only (rejected by `POST /v1/admin/login`). So the CRUD scenario is **env-gated** behind `LOCUST_ADMIN_*` credentials, and I added an always-on **customer-realm auth scenario** so the default `make perf-test` produces useful output with zero manual setup. Happy to reshape if you prefer a different split.
- Optional follow-up question: a dev-only admin seed script could make the admin scenario turnkey. There is an in-repo precedent (`_seed_real_admin` e2e fixture), but it bypasses the normal first-admin lifecycle (in-process DI, no setup gate), so I left it out pending your call.

## Notes
- Drift (deferred): `.claude/rules/commands.md` does not yet list `make perf-test`. Deliberately left out to keep this PR non-governor-changing — happy to include it or leave it for a follow-up `/sync-guidelines`, your preference.
- Pre-existing test failure (unrelated): `tests/unit/agents_shared/test_antigravity_hardening.py::test_native_write_file_edit_is_formatted` fails identically on current `main` in my environment (formatting-hook subprocess appears environment-dependent). Not touched by this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help setup quickstart demo demo-rag dev worker langfuse-env observability-langfuse observability-langfuse-down test lint format check check-core check-full check-minimal smoke-examples clean diagrams
+.PHONY: help setup quickstart demo demo-rag dev worker langfuse-env observability-langfuse observability-langfuse-down test lint format check check-core check-full check-minimal smoke-examples perf-test clean diagrams
 
 LANGFUSE_ENV_FILE := _env/langfuse.env
 MINIMAL_UV_ENV := /tmp/fastapi-agent-blueprint-minimal-venv
@@ -107,6 +107,26 @@ test-dynamo:
 ## Run tests with coverage
 test-cov:
 	uv run pytest tests/ -v --cov=src --cov-report=term-missing
+
+# Locust performance-test defaults — override per run, e.g.
+#   make perf-test PERF_USERS=50 PERF_RUN_TIME=2m
+PERF_HOST ?= http://127.0.0.1:8001
+PERF_USERS ?= 10
+PERF_SPAWN_RATE ?= 2
+PERF_RUN_TIME ?= 30s
+
+# Admin CRUD scenario needs LOCUST_ADMIN_USERNAME / LOCUST_ADMIN_PASSWORD —
+# see docs/operations/performance-locust.md. Numbers are illustrative only.
+## Headless Locust perf run against a local server (start `make quickstart` first)
+perf-test:
+	uv run locust \
+		-f tests/perf/locustfile.py \
+		--headless \
+		--host $(PERF_HOST) \
+		--users $(PERF_USERS) \
+		--spawn-rate $(PERF_SPAWN_RATE) \
+		--run-time $(PERF_RUN_TIME) \
+		--stop-timeout 10
 
 ## Run linter
 lint:
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,6 +34,7 @@ Quick index for the `docs/` folder.
 |---|---|
 | [operations/observability-otel.md](operations/observability-otel.md) | OpenTelemetry — Jaeger / Tempo / Phoenix setup |
 | [operations/observability-langfuse.md](operations/observability-langfuse.md) | Langfuse opt-in for LLM prompt tracing |
+| [operations/performance-locust.md](operations/performance-locust.md) | Locust performance-test harness — `make perf-test`, scenarios, reading output |
 
 ## Architecture decisions
 
--- a/docs/operations/performance-locust.md
+++ b/docs/operations/performance-locust.md
@@ -1,0 +1,167 @@
+# Locust Performance Testing — Operations Recipe
+
+This document covers the repo's Locust harness: how to run it against the
+zero-config quickstart server, how to read the output, and how to enable the
+optional admin CRUD scenario.
+
+This is a ready-to-adapt harness for a blueprint, not benchmarking
+infrastructure. Any numbers it produces are illustrative for the local
+quickstart setup (SQLite + InMemory broker on one machine) — treat them as a
+smoke reference, not production guarantees. Adopters should rerun the harness
+against their own infrastructure. It is intentionally **not** wired into CI.
+
+Never point the harness at a shared or production environment: the customer
+scenario registers accounts and the admin scenario creates and deletes users.
+
+## What you get
+
+`tests/perf/locustfile.py` defines three scenarios:
+
+| Scenario | User class | Requires | What it measures |
+|---|---|---|---|
+| Customer auth flow | `CustomerAuthUser` | nothing | `POST /v1/auth/register` (once per virtual user), `GET /v1/auth/me`, `POST /v1/auth/refresh`, and `POST /v1/auth/logout` on stop — all on the customer JWT realm |
+| Concurrent health reads | `HealthCheckUser` | nothing | `GET /health` (no-op app baseline) and `GET /health/db` (connection-pool acquisition + a real `SELECT 1`; 503 when the DB is down) |
+| Admin CRUD flow | `AdminCrudUser` | `LOCUST_ADMIN_*` env vars | `POST /v1/admin/login`, paginated `GET /v1/users`, and a create → read → update → delete cycle on `/v1/user` |
+
+`/health` and `/health/db` are tracked as separate stat rows on purpose — the
+first isolates app/HTTP responsiveness, the second adds a database roundtrip.
+The per-ID CRUD requests are grouped under one `/v1/user/[id]` row so each
+generated ID does not create its own statistics line.
+
+## Run it
+
+```bash
+make quickstart      # terminal 1 — zero-config server on 127.0.0.1:8001
+make perf-test       # terminal 2 — 30s headless run, summary printed on exit
+```
+
+Without admin credentials the run covers the customer + health scenarios and
+logs one line noting that the admin scenario is disabled. That default run is
+the supported zero-setup path.
+
+### Configuration
+
+| Variable | Default | Meaning |
+|---|---|---|
+| `PERF_HOST` | `http://127.0.0.1:8001` | Target base URL |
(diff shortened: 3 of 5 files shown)


Mr-DooSun:
Thanks for the thorough writeup, @Ayushhh26 — and for surfacing the realm-gate question instead of forcing a shape onto it. Your read is exactly right, so let me settle the open items.

**Scenario shape — keep it as is.** All `/v1/user` routes are admin-realm-gated (#199/#218) and the bootstrap admin is setup-only, so a turnkey always-on CRUD scenario isn't possible without a shortcut. Env-gating the admin CRUD behind `LOCUST_ADMIN_*` and adding an always-on customer-realm auth flow for the zero-setup default is the correct call. The `abstract = not ADMIN_SCENARIO_ENABLED` gate and splitting `/health` vs `/health/db` into separate stat rows are both nicely done.

**No dev-only admin seed script.** You already named the tradeoff and I agree — leave it out. Seeding a real admin in-process bypasses the first-admin setup lifecycle (temp password + forced change), and the blueprint shouldn't model that shortcut. The documented one-time dashboard provisioning is the right path.

**One change before merge — the admin scenario leaks a session.** `CustomerAuthUser` carefully revokes its refresh token in `on_stop`, but `AdminCrudUser` keeps only the access token, so each admin-enabled run leaves a live admin refresh-token session in the DB (it's persisted on login, 7-day TTL by default). Mirror the customer flow:

```python
# tests/perf/locustfile.py — AdminCrudUser
# before (end of on_start):
            self._access_token = response.json()["data"]["accessToken"]

# after:
            data = response.json()["data"]
            self._access_token = data["accessToken"]
            self._refresh_token = data["refreshToken"]

    def on_stop(self) -> None:
        # Mirror CustomerAuthUser: revoke the admin refresh token so opt-in
        # runs don't leave live admin sessions behind (on_stop also runs when
        # on_start failed, hence the guard).
        refresh_token = getattr(self, "_refresh_token", None)
        if refresh_token:
            self.client.post("/v1/admin/logout", json={"refreshToken": refresh_token})
```

Everything else lines up against the current API — endpoints, the camelCase request/response fields, the 200 success status, the `/v1/users` list vs `/v1/user/{id}` split, and the 503-on-DB-down note are all accurate.

**Don't touch `.claude/rules/commands.md`.** You were right to leave `make perf-test` out to keep this non-governor-changing — I'll fold the command/env reference into a separate `/sync-guidelines` pass on my side. Nothing for you to do there.

**The `test_antigravity_hardening` failure is unrelated** — thanks for flagging it; it's a known environment-dependent case in our harness suite that I'm tracking separately, not a blocker here.

Once the admin `on_stop` is in, this is good to merge. Really clean work — thanks for the patience while I got to it.

Ayushhh26:
Thanks for the review @Mr-DooSun 

I've added the admin logout fix in af0127d. I tested it by setting up a real admin locally and running the harness with the admin scenario on. Everything passed with no failures, and I checked the database afterwards to make sure the admin sessions actually get revoked now. They do. Good catch on that one.

Mr-DooSun (dismissed):
Reviewed against the shared Review Protocol (CORR / REG / STAB / CONTRACT / ARCH / SEC / GOV). **Verdict: PASS — merge-ready.**

No `src/` runtime change; contract, correctness, stability, security, and regression all check out against the current routers/schemas and CI config. Thanks for the careful work — the endpoint shapes, camelCase aliasing, 200 status codes, the 20-char username bound, the `pageSize` alias, and the `abstract`-based scenario gating all match the codebase exactly.

A few notes, none blocking:

**Scenario shape (your question) — accept as-is.** Your read is correct: all `/v1/user` routes are admin-realm-gated (#199/#218) and the bootstrap admin is setup-only, so the original "auth + CRUD on `/v1/user`" cannot be a single always-on flow. Env-gating the admin CRUD behind `LOCUST_ADMIN_*` and adding an always-on customer-realm auth scenario so the default `make perf-test` stays zero-setup is the right adaptation. No reshaping needed.

**`commands.md` drift — leave it to us.** You were right to keep `make perf-test` out of `.claude/rules/commands.md` to keep this PR non-governor-changing. Editing that path is itself a governed change, so a maintainer will fold the line in via a separate `/sync-guidelines` pass. Nothing for you to do here.

**Admin seed script — out of scope.** The dev-only first-admin seed idea is reasonable but bypasses the normal first-admin lifecycle; better tracked as a separate issue than expanded into this PR.

Verified specifics:
- **Contract** — `/v1/auth/*`, `/v1/admin/*`, `/v1/users` (`page` + `pageSize` alias), and `/v1/user/{id}` all match the routers, and `data.accessToken` / `data.refreshToken` / `data.id` match `SuccessResponse[...]` by-alias serialization.
- **Regression** — the module-level `import locust` is safe: the CI `test` job installs the `dev` group (which includes `locust`) before `pytest tests/`, and the `minimal-install` job runs only its single test file, so `tests/perf/` is never collected in a locust-less environment.
- **Security** — credentials are read from the environment only and never logged, and `_error_code` deliberately avoids echoing response bodies (dev-mode error details can carry submitted values).

Nice, self-contained contribution. 🙌

Mr-DooSun (commented):
Reviewed against the shared Review Protocol (CORR / REG / STAB / CONTRACT / ARCH / SEC / GOV). **Verdict: PASS — merge-ready.**

No `src/` runtime change; contract, correctness, stability, security, and regression all check out against the current routers/schemas and CI config. Thanks for the careful work — the endpoint shapes, camelCase aliasing, 200 status codes, the 20-char username bound, the `pageSize` alias, and the `abstract`-based scenario gating all match the codebase exactly.

A few notes, none blocking:

**Scenario shape (your question) — accept as-is.** Your read is correct: all `/v1/user` routes are admin-realm-gated (#199/#218) and the bootstrap admin is setup-only, so the original "auth + CRUD on `/v1/user`" cannot be a single always-on flow. Env-gating the admin CRUD behind `LOCUST_ADMIN_*` and adding an always-on customer-realm auth scenario so the default `make perf-test` stays zero-setup is the right adaptation. No reshaping needed.

**`commands.md` drift — leave it to us.** You were right to keep `make perf-test` out of `.claude/rules/commands.md` to keep this PR non-governor-changing. Editing that path is itself a governed change, so a maintainer will fold the line in via a separate `/sync-guidelines` pass. Nothing for you to do here.

**Admin seed script — out of scope.** The dev-only first-admin seed idea is reasonable but bypasses the normal first-admin lifecycle; better tracked as a separate issue than expanded into this PR.

Verified specifics:
- **Contract** — `/v1/auth/*`, `/v1/admin/*`, `/v1/users` (`page` + `pageSize` alias), and `/v1/user/{id}` all match the routers, and `data.accessToken` / `data.refreshToken` / `data.id` match `SuccessResponse[...]` by-alias serialization.
- **Regression** — the module-level `import locust` is safe: the CI `test` job installs the `dev` group (which includes `locust`) before `pytest tests/`, and the `minimal-install` job runs only its single test file, so `tests/perf/` is never collected in a locust-less environment.
- **Security** — credentials are read from the environment only and never logged, and `_error_code` deliberately avoids echoing response bodies (dev-mode error details can carry submitted values).

Nice, self-contained contribution. 🙌

PR #295, opened 2026-07-17
fix(examples): bound ChatReply.confidence to [0, 1] across chatbot examples

### LINKED ISSUE #294: Bound `ChatReply.confidence` to [0, 1] across the chatbot examples
Discovered during the #287 (`web_search_chatbot`) review — out of scope for that PR, tracking separately.

## Problem

In the chatbot examples, the PydanticAI agent's structured-output schema `ChatReply.confidence` has **no numeric constraint** — only a description:

```python
# examples/*/domain/dtos/chatbot_dto.py
confidence: float = Field(..., description="...between 0.0 and 1.0")
```

…while the API response schema enforces the range:

```python
# examples/*/interface/server/schemas/chatbot_schema.py
confidence: float = Field(..., ge=0.0, le=1.0, ...)
```

The service **persists the message before the router builds the response**:

`domain/services/chatbot_service.py` (`insert_data(...)`) → `interface/server/routers/chatbot_router.py` (`ChatResponse(...)`)

So if the LLM returns a confidence outside `[0, 1]` (e.g. `1.2`), the DB row is written, then `ChatResponse(...)` construction raises a pydantic `ValidationError` → the generic handler (`src/_core/exceptions/exception_handlers.py`) returns **500**. A client retry then creates a **duplicate persisted turn**.

## Affected examples

All share the same shape copied from `simple_chatbot` (#249):

- `examples/simple_chatbot/`
- `examples/chatbot_with_memory/`
- `examples/chatbot_with_guardrails/`
- `examples/web_search_chatbot/` (#287)

## Proposed fix

Add `ge=0.0, le=1.0` to `ChatReply.confidence` in each example's `domain/dtos/chatbot_dto.py`. This moves the failure to the agent's **output-validation stage** (PydanticAI re-prompts on output-schema failure) **before** persistence — no orphan row, no post-persist 500.

Add a unit test asserting an out-of-range model confidence is rejected before the DB insert.

## Scope notes

- Do this as **one sweep across all sibling examples** so they stay pattern-consistent (the examples deliberately mirror each other).
- `examples/`-only, no `src/` runtime change expected. Follow the copy-flow rules (no absolute `examples.*` imports).
- Not a blocker for #287 — that PR faithfully mirrors the existing `simple_chatbot` pattern; this is a pre-existing latent issue across the set.

Mr-DooSun:
## Problem

The PydanticAI agent output schema `ChatReply.confidence` had **no numeric constraint** (only a description), while the API response schema enforced `ge=0.0, le=1.0`. Because the service **persists the turn before the router builds the response**, a model confidence outside `[0, 1]` (e.g. `1.2`) wrote the DB row first, then failed `ChatResponse(...)` construction with a pydantic `ValidationError` → generic handler returns **500**. A client retry then created a **duplicate persisted turn**.

`chatbot_with_memory` was a worse instance: it persists **two** rows (user + assistant) before building the response, so an out-of-range confidence orphaned **two** rows.

## Fix

Add `ge=0.0, le=1.0` to `ChatReply.confidence` in all four chatbot examples, so an out-of-range value is rejected at the agent's **output-validation stage** — before persistence. This eliminates the orphan row(s) and the duplicate-on-retry. (If the model exhausts PydanticAI's output retries it still surfaces a 500, but *before* any persistence, so a retry is safe.)

Also bound `chatbot_with_memory`'s `ConversationTurnDTO.confidence` carrier as defense-in-depth.

## Affected examples

- `examples/simple_chatbot/`
- `examples/chatbot_with_memory/`
- `examples/chatbot_with_guardrails/`
- `examples/web_search_chatbot/`

## Tests

- Per-example unit test asserting boundary values (`0.0`, `1.0`) are accepted and out-of-range values (`1.2`, `-0.1`) raise `ValidationError`. The memory test also covers `ConversationTurnDTO` (both bounds).
- Per-example end-to-end regression test (cross-review follow-up): an invalid `FunctionModel` (`confidence=1.2`) wired through the **real adapter + service** raises `UnexpectedModelBehavior` at output validation and performs **zero inserts** (memory example: neither the user nor the assistant row) — pinning the #294 mechanism (`output_type` wiring + validate-before-persist order) against regression.

## Verification

- `pytest` for all four example suites — pass, including the `TestModel`/`FunctionModel` real-agent paths (confirming the new constraint doesn't break agent output generation)
- `tools/check_examples_copyflow.py` — 0 violations
- `ruff format` + `ruff check` + full `pre-commit` on changed files — pass

`examples/`-only; no `src/` runtime change.

Closes #294.

### DIFF
--- a/examples/chatbot_with_guardrails/domain/dtos/chatbot_dto.py
+++ b/examples/chatbot_with_guardrails/domain/dtos/chatbot_dto.py
@@ -8,7 +8,10 @@ class ChatReply(BaseModel):
 
     reply: str = Field(..., description="The assistant response text")
     confidence: float = Field(
-        ..., description="The confidence score of the reply, between 0.0 and 1.0"
+        ...,
+        ge=0.0,
+        le=1.0,
+        description="The confidence score of the reply, between 0.0 and 1.0",
     )
 
 
--- a/examples/chatbot_with_memory/domain/dtos/chatbot_memory_dto.py
+++ b/examples/chatbot_with_memory/domain/dtos/chatbot_memory_dto.py
@@ -8,7 +8,10 @@ class ChatReply(BaseModel):
 
     reply: str = Field(..., description="The assistant response text")
     confidence: float = Field(
-        ..., description="The confidence score of the reply, between 0.0 and 1.0"
+        ...,
+        ge=0.0,
+        le=1.0,
+        description="The confidence score of the reply, between 0.0 and 1.0",
     )
 
 
@@ -40,6 +43,6 @@ class ConversationTurnDTO(BaseModel):
     session_id: str
     user_message: str
     assistant_reply: str
-    confidence: float
+    confidence: float = Field(..., ge=0.0, le=1.0)
     tokens_used: int
     created_at: datetime
--- a/examples/simple_chatbot/domain/dtos/chatbot_dto.py
+++ b/examples/simple_chatbot/domain/dtos/chatbot_dto.py
@@ -8,7 +8,10 @@ class ChatReply(BaseModel):
 
     reply: str = Field(..., description="The assistant response text")
     confidence: float = Field(
-        ..., description="The confidence score of the reply, between 0.0 and 1.0"
+        ...,
+        ge=0.0,
+        le=1.0,
+        description="The confidence score of the reply, between 0.0 and 1.0",
     )
 
 
--- a/examples/web_search_chatbot/domain/dtos/chatbot_dto.py
+++ b/examples/web_search_chatbot/domain/dtos/chatbot_dto.py
@@ -8,7 +8,10 @@ class ChatReply(BaseModel):
 
     reply: str = Field(..., description="The assistant response text")
     confidence: float = Field(
-        ..., description="The confidence score of the reply, between 0.0 and 1.0"
+        ...,
+        ge=0.0,
+        le=1.0,
+        description="The confidence score of the reply, between 0.0 and 1.0",
     )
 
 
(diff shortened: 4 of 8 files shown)


Mr-DooSun:
## Review (Quality Gate — `/review-pr`)

**Scope**
- 4 example DTOs (`domain/dtos/*.py`) + 4 test files, `examples/` and `tests/` only, no `src/` change.
- Closes #294.

**Effect Answer**
- Adds `ge=0.0, le=1.0` to `ChatReply.confidence` (the PydanticAI agent output schema) in all four chatbot examples, so an out-of-range model confidence now fails at output-validation time — before the service persists the turn. This matches the API response schema's existing bound and removes the orphan-row / duplicate-on-retry failure mode described in #294.

**Verification performed**
- Read `domain/services/*.py` for all four examples: confirmed `chat_reply.confidence` flows unbounded from the agent into the persisted DTO and the router response in every case — matches the PR's root-cause description.
- Confirmed the response schemas (`interface/server/schemas/*.py`) already enforce `ge=0.0, le=1.0`, so this change closes the asymmetry rather than introducing a new constraint.
- Ran the full test suite for all four examples in an isolated worktree with the `pydantic-ai` extra installed (real adapter path, not just the stub): **27 passed**, including the new `test_out_of_range_confidence_fails_before_persistence` regression tests — each confirms `UnexpectedModelBehavior` is raised and the repository's insert is never called.
- `tools/check_examples_copyflow.py`: 0 violations.
- `ruff check` / `ruff format --check` on all 8 changed files: pass.
- `gh pr checks 295`: Validate / architecture / branch-name / lint / minimal-install / test all pass.

**Findings**
- None open.

**Verdict**
- PASS — matches #294's stated scope and acceptance criteria exactly, no regressions found.

Posting as a comment (not an approval) since I'm the PR author — GitHub doesn't allow self-approval.

Issue #17, opened 2026-03-25
Add error notification (Slack, Discord)


Mr-DooSun:
### Problem

When errors occur in production, there is no automated notification mechanism to alert the team through common communication channels. Issues may go undetected until users report them.

### Proposed Solution

Add an error notification system:
- Webhook integration for Slack and Discord
- Configurable severity thresholds (e.g., only notify on 5xx errors or critical exceptions)
- Rate limiting to prevent notification storms
- Integration with the existing exception middleware for automatic triggering
- Channel routing (e.g., critical → #alerts, warnings → #monitoring)

### Area

Core (_core)

### Alternatives Considered

- External APM tools only (Sentry, Datadog): adds cost and external dependency — basic webhook notifications are complementary and free
- Email notifications: slower feedback loop than instant messaging

vraj2131:
Hi, Can I take this one?

Mr-DooSun:
@vraj2131 Sure, thanks for your interest! Please go ahead.

Before implementing, please share a brief approach first. The expected scope is:
- Slack/Discord webhook notification support
- configurable severity threshold
- basic rate limiting
- integration with the existing exception middleware
- tests for notification triggering and suppression behavior

vraj2131:
Thanks! Here is the plan before I start:

- I'll add it as an optional piece of core infra (Slack and Discord webhook adapters, plus a no op fallback when it's disabled), following the same Protocol and Selector pattern the project already uses for things like LLM and Storage. That way it's toggleable through env vars and costs nothing when it's off.

- Webhook calls will go through the existing HttpClient, so there's no new HTTP dependency.

- Severity threshold and rate limiting will just be a couple of new settings plus a small in memory cooldown. This is not the slowapi rate limiting item from the roadmap, since that's for HTTP endpoints. This one just stops repeat error spam.

- It will hook into the existing exception handlers (generic_exception_handler and custom_exception_handler). If a webhook call fails, it will just get logged and never turn into a second error for the user.

- Tests will cover threshold gating, suppression on repeated errors, and payload building for both Slack and Discord.

- This shouldn't need a new domain, a new DB table, or an ADR, so I'll keep it as small as possible.

Let me know if that lines up with what you had in mind, then I'll get started.

Mr-DooSun:
This plan lines up well — go ahead and get started. 🙌

[...]

PR #311, opened 2026-07-27
docs(operations): add error notification runbook (#307)

Mr-DooSun:
## Summary

Adds the missing operator runbook for the error-notification webhook feature
(#17 / PR #304). It was the only optional infra with non-trivial setup and no
`docs/operations/` runbook — RDB migrations, OTEL, Langfuse, and the Locust
harness all have one.

PR #306 already covered the env templates and harness rules, and PR #308 (#305)
covered the reviewer-facing `security-checklist.md` §13 items. This is the
operator-facing counterpart, and the last piece of the #17 documentation set.

## What the runbook covers

The `_env/*.example` comments already say *which* variables exist. These are the
behaviours only discoverable by reading `error_notifier.py` and
`exception_handlers.py`:

- **The exact dispatch surface** — three call sites, and the message format each
  sends (`str(exc)` for `BaseCustomException`, `f"{type(exc).__name__}: {exc}"`
  for anything reaching `generic_exception_handler`).
- **What does NOT alert** — 422 validation errors and `raise HTTPException(500)`
  never dispatch; Taskiq worker task failures and NiceGUI admin exceptions have
  no notifier wired at all (now tracked in #310).
- **The per-process cooldown caveat** — `ErrorNotifier` is a per-process
  Singleton holding an in-memory dict keyed by `error_code`, so uvicorn workers
  x replicas each get their own window. Expect up to N alerts per window, not 1.
- **The disabled path** — `notification_client_disabled` fires at the *first
  dispatch attempt*, not at boot (lazily-resolved Singleton), and
  `notification_suppressed` writes the full alert text to the log at `INFO`.
- **The un-redacted payload risk**, split per handler rather than stated flatly.
- **A local-sink verification recipe** needing no Slack/Discord account.
- **A troubleshooting table**.

## Corrections to the issue body

Three claims in #307 did not survive verification against the running code. The
runbook documents the corrected behaviour; details in the issue comment.

1. "N server/worker processes" — the notifier is server-only; worker and admin
   failures never alert. The cooldown multiplier is uvicorn workers x replicas.
2. "the one-time `notification_client_disabled` **startup** warning" — it is not
   a startup line. Measured: 0 occurrences at boot, 1 after the first gated error.
3. "unhandled `IntegrityError` renders the statement and bound parameters" — true
   for the raw path, but standard repository access goes through
   `Database.session()`, which re-raises a curated `400 [DB_INTEGRITY_ERROR]`
   (below the default threshold, no SQL). The raw path is reachable where an
[...]

Issue #310, opened 2026-07-27
Error notifications: worker task failures and admin exceptions are never alerted


Mr-DooSun:
### Problem

Error notifications (#17 / PR #304) are wired **only** into the FastAPI global
exception handlers. `ErrorNotifier.maybe_dispatch` has exactly three call sites,
all in `src/_core/exceptions/exception_handlers.py`:

- `custom_exception_handler` — any `BaseCustomException`
- `generic_exception_handler` — mapped LLM/provider error
- `generic_exception_handler` — unhandled exception

Nothing else dispatches. Two production failure surfaces therefore produce
**no alert at all**, no matter how `NOTIFICATION_*` is configured:

| Surface | Current behaviour |
|---|---|
| Taskiq worker task failures | `src/_apps/worker/bootstrap.py::_install_middleware` installs task-context logging + retry middleware only — no notifier is wired |
| NiceGUI admin page / callback exceptions | `handle_uncaught_admin_exception` (`src/_core/infrastructure/admin/error_handler.py`) is log-only by design |

This is easy to misread as a bug in the operator's configuration: the server
alerts correctly, so the pipeline is visibly "working", while a repeatedly
failing background job stays silent.

Documented as a coverage limit in the new runbook
(`docs/operations/error-notifications.md` § "What does NOT get alerted", #307),
so this issue is about deciding the behaviour, not about disclosure.

### Proposed Solution

Decide and record which of these is intended:

1. **Extend coverage to the worker** — a Taskiq middleware `on_error` hook that
   resolves `error_notifier` from the worker container and calls
   `maybe_dispatch`. Notes:
   - `ErrorNotifier` is a per-process Singleton, so the worker gets its own
     cooldown window — consistent with the existing server behaviour
   - the worker has no HTTP status; a synthetic severity would be needed for the
     `NOTIFICATION_SEVERITY_THRESHOLD` comparison, or the threshold check would
     have to be bypassed for task failures
   - the retry middleware means a transient failure would otherwise alert on
     every attempt — dispatch should probably fire only on permanent failure /
     final attempt
2. **Keep it HTTP-only by design** — worker and admin failures are the log
   aggregator's job. In that case the decision belongs in the runbook and in
   `AGENTS.md` § Optional Infrastructure Toggles as an explicit non-goal, rather
   than being implicit in the call-site list.

Option 2 is defensible: it keeps the feature small and avoids a second severity
model. But it should be a stated decision, because the current state reads as an
[...]

PR #313, opened 2026-07-28
feat(_core): add severity-based channel routing for error notificatio…

### LINKED ISSUE #286: Add channel routing for error notifications (Slack, Discord)
### Problem
The base error-notification feature (#17) sends Slack/Discord webhooks on
qualifying exceptions, but routes everything to a single configured target.
Teams usually want severity-based fan-out so noisy warnings and true incidents
don't land in the same channel.

### Proposed Solution
Build on top of the notification infra from #17:
- Route by severity, e.g. `critical → #alerts`, `warnings → #monitoring`
- Configurable mapping via settings/env (keep the single-target default when unset)
- Reuse the existing Protocol + Selector adapters and the severity threshold
  already introduced in #17 — this is routing on top, not a new transport

### Notes
- Deferred from #17 to keep the base notification PR focused.
- No new domain/table expected; likely just settings + a small routing layer
  in front of the existing webhook adapter(s).

### Area
Core (_core)

siddhanttiwari19:
## Related Issue
- Fixes #286

## Change Summary
- Add severity-based channel routing for error notifications, on top of #17's single-target `ErrorNotifier`.
- New `NotificationRouter` (`src/_core/infrastructure/notification/notification_router.py`) resolves a critical-tier vs. warning-tier webhook client per status code.
- `ErrorNotifier` takes an optional `notification_router` param (default `None`). Unset, behavior is byte-for-byte identical to #17 — verified by leaving all 7 pre-existing `ErrorNotifier` tests unmodified and green.
- New settings: `NOTIFICATION_WARNING_THRESHOLD` (must be `< NOTIFICATION_SEVERITY_THRESHOLD`, unset by default → warning tier disabled entirely), `NOTIFICATION_CRITICAL_WEBHOOK_URL`, `NOTIFICATION_WARNING_WEBHOOK_URL` (each falls back to the existing single-target webhook when unset).
- `CoreContainer` wires two new `Selector` providers (critical/warning) reusing the existing `_build_notification_client` factory — same provider (Slack or Discord), different webhook URL per tier. No new transport.
- Docs: `AGENTS.md` § Optional Infrastructure Toggles new row; `_env/local.env.example` example block.

## Type of Change
- [x] feat: New feature

## Checklist
- [x] Tests pass (`552 passed`, core suite; `178` notification-scoped, `25` new / `0` regressions)
- [x] Linting passes (`ruff check src/`, `ruff format --check src/`)

## How to Test
- `NOTIFICATION_WARNING_THRESHOLD` unset (default): behavior identical to #17 — only `NOTIFICATION_SEVERITY_THRESHOLD`-and-above notifies, single target.
- Set `NOTIFICATION_WARNING_THRESHOLD=400` with `NOTIFICATION_CRITICAL_WEBHOOK_URL` / `NOTIFICATION_WARNING_WEBHOOK_URL` pointing at two different Slack channels: a 404 routes to the warning channel, a 500 routes to the critical channel.
- `uv run pytest tests/unit/_core/infrastructure/notification/ tests/unit/_core/test_config.py tests/unit/_core/infrastructure/di/test_core_container_selectors.py tests/integration/_core/infrastructure/test_optional_infra.py -v`

---

## Governor Footer
- trigger: yes
- reviewer: human:Mr-DooSun
- rounds: 3
- r-points-fixed: 3
- r-points-deferred: 3
- r-points-rejected: 0
- pr-scope-notes: two-tier (critical/warning) routing only, single provider per tier via webhook-URL override, not multi-provider fan-out; matches issue #286 scope. Governor-changing via the AGENTS.md Optional Infrastructure Toggles row. 3 R-points fixed in-PR (two container-wiring defects, one cross-tier cooldown regression); 3 deferred with rationale and re-homed to tracked issues #315 / #316 / #317 rather than a 4th review round.
- final-verdict: merge-ready
- links: https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/313, https://github.com/Mr-DooSun/fastapi-agent-blueprint/issues/286, https://github.com/Mr-DooSun/fastapi-agent-blueprint/issues/315, https://github.com/Mr-DooSun/fastapi-agent-blueprint/issues/316, https://github.com/Mr-DooSun/fastapi-agent-blueprint/issues/317


src/_core/infrastructure/di/core_container.py
```diff
@@ -409,9 +439,47 @@ class CoreContainer(containers.DeclarativeContainer):
...
+
+    notification_critical_client = providers.Selector(
+        _notification_critical_selector,
+        enabled=providers.Singleton(
+            _build_notification_client,
+            http_client=http_client,
+            provider=settings.notification_provider,
+            webhook_url=settings.notification_critical_target,
+        ),
+        disabled=providers.Singleton(_build_noop_notification_client),
```
Mr-DooSun: **`[OPEN][HIGH][REG]`** — finding key: `REG | runtime-evidence | src/_core/infrastructure/di/core_container.py:457 | noop-client-warning-emitted-three-times`

**basis:** runtime-evidence + contract-evidence

`NoopNotificationClient` emits its warning from `__init__` ([`noop_notification_client.py:17`](https://github.com/Mr-DooSun/fastapi-agent-blueprint/blob/main/src/_core/infrastructure/notification/noop_notification_client.py#L17)), so instance count equals log-line count. This line, plus `:439` and `:468`, each declare a **distinct** `providers.Singleton(_build_noop_notification_client)`. On the disabled path `notification_router` (`:471`) eagerly resolves both new tiers while `error_notifier` (`:481`) still resolves the base client — three providers, three instances, three warnings.

Probe, resolving `error_notifier()` exactly once with every `NOTIFICATION_*` setting unset:

```text
main   @ 1e301cf → notification_client_disabled ×1
pr/313 @ 0ce9b18 → notification_client_disabled ×3
```

**Impact:** this lands on the **default, un-opted-in path** — every deployment without `NOTIFICATION_PROVIDER`, including `make quickstart` — so it escapes the opt-in boundary the rest of the feature respects. It contradicts `docs/operations/error-notifications.md:214`, which documents the line as `once per process, at the first dispatch attempt`, and that runbook has a troubleshooting row keyed on it. Nothing currently pins the emission count, which is why the suite stayed green.

**Recommended fix** — declare the Noop provider once and share it. Verified locally: 3 → 1 warnings, `76 passed` across `tests/unit/_core/infrastructure/notification/`, `tests/unit/_core/infrastructure/di/`, and `tests/integration/_core/infrastructure/test_optional_infra.py`.

```python
# before — three independent Singletons at :439, :457, :468
    disabled=providers.Singleton(_build_noop_notification_client),
```

```python
# after — one shared Singleton declared just above `notification_client`,
#         then referenced from all three Selectors
    _noop_notification_client = providers.Singleton(_build_noop_notification_client)

    notification_client = providers.Selector(
        _notification_selector,
        enabled=providers.Singleton(...),
        disabled=_noop_notification_client,
    )
    # ... and `disabled=_noop_notification_client` in the two new Selectors as well
```

Please also add a regression test asserting a single `notification_client_disabled` emission, so this cannot silently return.

Note: the `enabled` branches are correct as written — **do not** collapse those into a shared provider. Each tier legitimately needs its own client because the webhook URL differs per tier.

src/_core/infrastructure/notification/error_notifier.py
```diff
@@ -52,9 +65,28 @@ def _should_notify(self, status_code: int, error_code: str) -> bool:
...
+            return min(self._severity_threshold, self._router.warning_threshold)
+        return self._severity_threshold
+
+    async def _safe_send(self, status_code: int, error_code: str, message: str) -> None:
         try:
-            await self._client.send(message)
+            client = (
+                self._router.resolve(status_code)
+                if self._router is not None
+                else self._client
```
Mr-DooSun, line 103: **`[OPEN][MEDIUM][REG]`** — finding key: `REG | diff-evidence | src/_core/infrastructure/notification/error_notifier.py:85 | notification-client-unreachable-in-safe-send`

**basis:** diff-evidence

`core_container.py:484` passes `notification_router=notification_router` unconditionally, so `self._router` is never `None` in production and this `else self._client` branch is unreachable outside unit tests. `notification_client` (`core_container.py:481`) is still constructed on every path but never sends anything.

**Impact:** the documented single-target client becomes dead wiring. On the disabled path it is the third `NoopNotificationClient` behind the `HIGH` finding; on the enabled path it builds a third live Slack/Discord adapter that never posts. It also falsifies `docs/ai/shared/security-checklist.md:419`, which records `_build_notification_client` … `settings.notification_webhook_url` as "the sole production construction site" — that grep is what a reviewer follows to audit where the webhook credential actually flows, so it needs to stay true.

**Recommended fix** — make the injection match the intent. Either direction is fine; pick whichever reads better to you.

```python
# before — core_container.py:479
    error_notifier = providers.Singleton(
        _build_error_notifier,
        notification_client=notification_client,
        severity_threshold=settings.notification_severity_threshold,
        cooldown_seconds=settings.notification_cooldown_seconds,
        notification_router=notification_router,
    )
```

```python
# after — option A: inject the router only when routing is actually configured,
#         so the documented single-target path stays the real one
    notification_router = providers.Selector(
        _notification_routing_selector,   # "enabled" iff notification_warning_threshold is set
        enabled=providers.Singleton(_build_notification_router, ...),
        disabled=providers.Object(None),
    )
    # `error_notifier` unchanged — the `else self._client` branch becomes live again

# after — option B: keep the router always-on and drop the now-unused base client
#         from `_build_error_notifier` and `ErrorNotifier.__init__`
```

Option A keeps `ErrorNotifier`'s two-mode contract meaningful and keeps the 7 pre-existing tests exercising a path production actually takes. Option B is less code but means the class no longer has a single-target mode at all — in which case the `notification_router: NotificationRouter | None = None` default and its docstring should stop describing one.

Mr-DooSun (dismissed):
## Review — PR #313 (round 1)

Thanks for this — it is a well-built contribution. The `Selector` + lazy-factory wiring follows ADR 042 correctly, the docstrings name their own tradeoffs instead of hiding them, boot-time threshold-ordering validation is there, and the 25 new tests are genuinely useful. The design is sound; everything below is one spot in the container wiring.

**Verdict: FAIL** — one `OPEN` `HIGH` `REG` finding contradicts the PR's stated intent. Two code asks, both small. **You do not need to touch any documentation** — the doc sync is maintainer work and is tracked in the ledger below.

### Effect Answer

The PR adds an opt-in two-tier severity router in front of #17's single-target `ErrorNotifier`: `NOTIFICATION_WARNING_THRESHOLD` widens `_should_notify`'s gate downward, and `NotificationRouter.resolve()` picks a critical vs. warning client per status code. Unset, only the critical tier ever resolves — that part is correct.

However, `CoreContainer` wires the router **unconditionally**, so the default (un-opted-in) path changes in two observable ways: three separate `NoopNotificationClient` instances are constructed instead of one, and the injected `notification_client` becomes unreachable inside `_safe_send`. Both are flagged inline.

The PR body states "Unset, behavior is byte-for-byte identical to #17 — verified by leaving all 7 pre-existing `ErrorNotifier` tests unmodified and green." That claim is true at the `ErrorNotifier` level, but the container never leaves `notification_router` unset, so the verified path is one production never takes. The 7 cited tests construct `ErrorNotifier` directly and bypass the container, which is why CI stayed green.

### Verification performed

Run against head `0ce9b18` in an isolated worktree:

| Check | Result |
|---|---|
| Full suite (`pytest tests/`) | **1542 passed, 32 skipped** |
| 10 DynamoDB integration errors | environmental (`dynamodb-local` not running) — reproduced identically on `main`, **not** caused by this PR |
| `ruff check` / `ruff format --check` | clean, 610 files |
| `mypy` on the 3 changed src files | clean |
| `tools/check_language_policy.py` | 0 violations / 228 files |
| `tools/check_governor_footer.py` | 0 violations |
| Router band logic (400/404/499/500/503) | correct in both `warning_threshold` states |
| Config validation env scope | runs for all envs, not `STRICT_ENVS`-gated — confirmed |
| Security checklist §13 | no new credential leak; the two new webhook URLs are never logged |

One correctness point worth recording as verified rather than left open: `_effective_min_threshold()` and `resolve()` **cannot** disagree. Settings validation enforces `warning_threshold < severity_threshold`, so the widened gate always lands in a resolvable tier and the cooldown is never burned on a `None` resolve. Good.

### Summary Finding Ledger

Out-of-diff finding — `docs/operations/error-notifications.md` is absent from this PR's diff, so GitHub cannot anchor a resolvable thread to it (see review-protocol §5).

- [ ] **`GOV | rule-source | docs/operations/error-notifications.md | runbook-not-synced-with-channel-routing`** — `[OPEN][MEDIUM][GOV]`
  **basis:** rule-source — `AGENTS.md` § Drift Management: "Update related docs in the same change when shared rules or harness behavior changes"
  **Impact:** the operator runbook added by PR #311 last week is untouched and is now wrong in four places — the env-var table (`:154`) omits all three new variables; the disabled-path row (`:214`, "once per process, at the **first dispatch attempt**") states a contract the inline `HIGH` finding breaks; the Related section (`:362`) still calls #286 "a follow-up"; and the worker troubleshooting row (`:348`) now has a second remedy. `docs/ai/shared/project-dna.md:624` likewise still ends "Channel routing is a follow-up (#286)." Both `.claude/rules/commands.md` and `security-checklist.md` §13 point operators and reviewers at this runbook as canonical.
  **Recommended fix:** maintainer-absorbed via `/sync-guidelines` after merge. **Not an ask for @siddhanttiwari19.**

Additional drift the same sync pass will close (recorded for the maintainer, not findings against this PR):

| target | reason |
|---|---|
| `.claude/rules/{project-status,commands,project-overview,architecture-conventions}.md` | #286 sits in "Not Yet Implemented"; Error Notification sections lack the three new variables |
| `docs/ai/shared/security-checklist.md` §13 | detection greps at `:418` / `:427` key off `slack_webhook_url\|discord_webhook_url` and `(?:SLACK\|DISCORD)_WEBHOOK_URL`, so the new `NOTIFICATION_{CRITICAL,WARNING}_WEBHOOK_URL` credentials are invisible to both |
| `docs/reference.md:295`, `README.md:113` | feature line and checkbox omit channel routing |
| `_env/quickstart.env.example:120` | carries a `NOTIFICATION_*` block; only `local.env.example` was updated |

### Questions (not findings — no action required to merge)

1. **Cross-tier cooldown masking.** `test_cooldown_is_shared_across_tiers_per_error_code` pins that a warning-tier 404 suppresses a subsequent critical-tier 500 sharing an `error_code`. I audited all 37 `error_code` / `status_code` pairs under `src/` — **none spans below-500 and 500-or-above**, so there is no reachable path today and this stays latent rather than a defect. Recording it because #310 deliberately scoped the worker cooldown key to `{task_name}:{error_code}` to prevent exactly this class of mute. Intentional here, or should the key become `{tier}:{error_code}`?

2. **`NOTIFICATION_WARNING_THRESHOLD` set alone.** With no `NOTIFICATION_WARNING_WEBHOOK_URL`, both tiers resolve to the same webhook, so 4xx floods the critical channel — the outcome #286 exists to prevent — and the `_env` example lists the threshold first. The `AGENTS.md` row documents the fallback as deliberate ("rather than dropping that tier"), so this is a design call, not a defect. Is a boot-time guard wanted instead?

3. **Unreachable config pinned by tests.** Both `test_enabled_via_explicit_override` cases set `slack_webhook_url=None` alongside a per-tier URL, but boot validation rejects that state (a set provider requires its matching base URL). Harmless, though it pins a configuration that cannot exist.

### Coverage

- `[OK][ARCH]` architecture-review-checklist §1 — no Domain → Infrastructure import; `NotificationRouter` lives in `infrastructure/notification/` and types against `BaseNotificationProtocol` from `_core/domain/protocols/`
- `[OK][ARCH]` §4 — `DeclarativeContainer` preserved; new providers use `Selector` + `Singleton` per ADR 042
- `[OK][ARCH]` §5 — 25 new tests across router unit, notifier integration, container selectors, config, and optional-infra
- `[OK][SEC]` security-checklist §13 — `_safe_send` still logs `exc_type` only; `_env` examples carry placeholders, not live URLs
- `[SKIP][CONTRACT]` no API schema, OpenAPI, status code, task payload, or migration touched

### Sync Required

`true` — closed by `/sync-guidelines` after merge, on the maintainer side.

src/_core/infrastructure/di/core_container.py
siddhanttiwari19: Fixed — shared a single _noop_notification_client Singleton across all three Selectors' disabled branches instead of one each. Verified: 1 warning now (was 3), same probe as your comment. Added a regression test pinning the count.

src/_core/infrastructure/notification/error_notifier.py
siddhanttiwari19: Went with option A — notification_router is now its own providers.Selector gated by notification_warning_threshold is not None, disabled=providers.Object(None). Unset, notification_router() returns None and ErrorNotifier's else self._client branch is live again. Kept the two-mode contract and docstring as-is since this restores what they already described. Updated test_notification_router_returns_router_instance (now asserts None by default) and added a selector-level test for the new gate.

src/_core/infrastructure/notification/error_notifier.py
```diff
@@ -38,12 +51,12 @@ def maybe_dispatch(
...
         if not self._should_notify(status_code, error_code):
             return
-        task = asyncio.create_task(self._safe_send(error_code, message))
+        task = asyncio.create_task(self._safe_send(status_code, error_code, message))
         self._background_tasks.add(task)
         task.add_done_callback(self._background_tasks.discard)
 
     def _should_notify(self, status_code: int, error_code: str) -> bool:
-        if status_code < self._severity_threshold:
+        if status_code < self._effective_min_threshold():
```
Mr-DooSun, line 59: **`[OPEN][HIGH][REG]`** — finding key: `REG | runtime-evidence | src/_core/infrastructure/notification/error_notifier.py:59 | warning-tier-4xx-mutes-critical-tier-5xx`

**basis:** runtime-evidence + test-evidence

This is the one thing I need before merge, and it is a case **round 1 got wrong** — it was recorded as a Question with the conclusion "latent, no reachable path today". The premise was incomplete, so re-raising it as a finding now.

Round 1 audited the 37 *declared* `error_code` / `status_code` pairs under `src/` and found none spanning the 500 boundary. But `BaseCustomException.__init__` takes **both** as per-instance arguments with defaults (`status_code: int = 400`, `error_code: str = "CUSTOM_ERROR"` — `src/_core/exceptions/base_exception.py:4-11`), so two raises can land in different tiers under one shared key without any subclass being declared.

Probe using only the shipped base class and its default `error_code`, with `NOTIFICATION_WARNING_THRESHOLD=400`:

```text
soft = BaseCustomException(status_code=404, message="profile not found")
hard = BaseCustomException(status_code=500, message="payment ledger corrupt")
same error_code?  True   ('CUSTOM_ERROR')

routing ON   → warning=['404 [CUSTOM_ERROR]: profile not found']   critical=[]     ← the 500 is SWALLOWED
routing OFF  → base   =['hard']                                                    ← the 500 is delivered (#17)
```

**Impact:** enabling channel routing can silently drop a 5xx incident alert for the whole `NOTIFICATION_COOLDOWN_SECONDS` window (default 60s). The `routing OFF` control line is the important half — this is a behavioural regression against #17, not just a new-feature quirk. Before this PR a sub-500 code never passed the gate, so it could never write a cooldown entry and could never consume a 5xx's slot. It also inverts #286's own goal ("so noisy warnings and true incidents don't land in the same channel" — here the noisy warning *deletes* the true incident), and it runs against the position #310 already settled for the worker path, where the key was deliberately scoped `{task_name}:{error_code}` because "a bare code would let one noisy task suppress every other task's alert" (`taskiq_middleware.py:14-17`).

Note that `test_error_notifier.py:231` (`test_cooldown_is_shared_across_tiers_per_error_code`) currently pins this as intended, asserting `critical.sent == []` at `:255` — so the fix needs that test's expectation inverted.

**Recommended fix** — scope the cooldown key by tier, mirroring #310. Verified locally: cross-tier mute gone, within-tier cooldown still suppresses, `#17` default path byte-identical; `53 passed` across `tests/unit/_core/infrastructure/notification/` and `tests/integration/_core/infrastructure/test_optional_infra.py`, with only `test_cooldown_is_shared_across_tiers_per_error_code` failing as expected.

```python
# before — error_notifier.py:58-66
    def _should_notify(self, status_code: int, error_code: str) -> bool:
        if status_code < self._effective_min_threshold():
            return False
        now = time.monotonic()
        last_notified = self._last_notified_at.get(error_code)
        if last_notified is not None and (now - last_notified) < self._cooldown_seconds:
            return False
        self._last_notified_at[error_code] = now
        return True
```

```python
# after
    def _should_notify(self, status_code: int, error_code: str) -> bool:
        if status_code < self._effective_min_threshold():
            return False
        key = self._cooldown_key(status_code, error_code)
        now = time.monotonic()
        last_notified = self._last_notified_at.get(key)
        if last_notified is not None and (now - last_notified) < self._cooldown_seconds:
            return False
        self._last_notified_at[key] = now
        return True

    def _cooldown_key(self, status_code: int, error_code: str) -> str:
        """Cooldown key, scoped by severity tier when routing is active (#286).

        Without a router the key is the bare ``error_code``, unchanged from
        #17. With one, a warning-tier 4xx and a critical-tier 5xx sharing an
        ``error_code`` must not share a quiet window — otherwise the 4xx
        silently mutes the incident alert. Mirrors #310's
        ``{task_name}:{error_code}`` scoping on the worker path.

        The band test matches ``NotificationRouter.resolve`` exactly, so the
        key and the delivered channel can never disagree.
        """
        if self._router is None:
            return error_code
        tier = "critical" if status_code >= self._severity_threshold else "warning"
        return f"{tier}:{error_code}"
```

Keeping the bare key when `self._router is None` is deliberate — please **do not** scope it unconditionally, or the 7 pre-existing `#17` tests stop exercising the path production actually takes on the default config.

Then invert `test_error_notifier.py:231` to assert both tiers receive (`test_cooldown_is_scoped_per_tier` reads better), and optionally add the within-tier case as a guard that the cooldown still works.

If you would rather not take another round on this, say so and I will absorb the fix on the maintainer side — your authorship on the feature stands either way.

Mr-DooSun (dismissed):
## Review — PR #313 (round 2)

Both round-1 findings are properly fixed, and I re-verified each by probe rather than taking the reply at its word — the shared Noop Singleton and the option-A router gate are exactly right, and the regression tests you added pin both. Thank you for the quick turnaround.

**Verdict: FAIL** — but please read the scope before it lands badly: **one** code ask, and it is a case round 1 got wrong, not something you introduced carelessly. The two ledger items below are **maintainer work and explicitly not asks for you** — one of them is our PR template being wrong, which is what failed CI on you.

### Effect Answer

On the default path `notification_router()` now returns `None`, `ErrorNotifier`'s `else self._client` branch is live again, and the disabled warning is back to **1 emission**, matching `main`. The enabled path also works end-to-end — with real env vars, `resolve(404)` routes to the warning channel, `resolve(500)` to the critical channel, `399` to neither.

The opt-in path does carry one reachable regression. Because the gate widens (`error_notifier.py:59`) while the cooldown stays keyed on the bare `error_code` (`:62`, `:65`), a warning-tier 4xx now consumes the quiet window of a critical-tier 5xx sharing that code. Round 1 recorded this as latent after auditing the 37 *declared* `error_code`/`status_code` pairs under `src/`. **That premise was incomplete** — `BaseCustomException` takes both as per-instance args with defaults (`status_code=400`, `error_code="CUSTOM_ERROR"`), so it is reachable with the shipped base class and no new code. Details inline.

### Verification performed

Isolated worktree at `69b95dd`, full `uv sync --group dev --extra admin --extra aws`.

| Check | Result |
|---|---|
| Full suite (`pytest tests/`) | **1545 passed, 32 skipped** |
| 10 DynamoDB integration errors | environmental (`aiohttp ClientConnectorError`) — reproduced identically on `main`, **not** caused by this PR |
| Notification-scoped suite | 196 passed |
| `ruff check` / `ruff format --check` | clean, 610 files |
| `pre-commit run --from-ref main --to-ref HEAD` | all Passed (incl. secret detection, Tier-1 language policy) |
| `tools/check_language_policy.py` | 0 violations / 228 files |
| CI | `test` / `lint` / `architecture` / `minimal-install` / `branch-name` pass; **`Validate` fails** (ledger item 2 — our bug, not yours) |

Round-1 fix probes, resolving `error_notifier()` exactly once with every `NOTIFICATION_*` setting unset:

```text
main   @ 1e301cf → notification_client_disabled ×1
pr/313 @ 69b95dd → notification_client_disabled ×1     (was ×3 at 0ce9b18)
                   notification_router() -> None
                   error_notifier._router -> None
                   _effective_min_threshold() -> 500
```

Enabled path, probed the way production reads config (env vars set **before** import — the gap you documented honestly in `test_optional_infra.py`, since container kwargs capture `settings.*` at class-import time):

```text
router.warning_threshold  -> 400
_effective_min_threshold() -> 400
resolve(404) -> SlackNotificationAdapter @ .../MONITORING
resolve(500) -> SlackNotificationAdapter @ .../ALERTS
resolve(399) -> None
dispatched   -> [('warning', 'soft'), ('critical', 'hard')]
```

One correctness point worth recording as verified rather than left open, unchanged from round 1: `_effective_min_threshold()` and `resolve()` cannot disagree, because boot validation enforces `warning_threshold < severity_threshold`. Good.

### Summary Finding Ledger

Round-1 keys, each re-verified against `69b95dd`:

- [x] ~~**`REG | runtime-evidence | src/_core/infrastructure/di/core_container.py:457 | noop-client-warning-emitted-three-times`**~~ **FIXED** — shared `_noop_notification_client` Singleton at `core_container.py:448`, referenced from all three Selectors' `disabled` branches (`:458`, `:476`, `:487`). Probe: **1** emission at head vs 3 at `0ce9b18`, matching `main`. `test_noop_notification_client_disabled_warning_emitted_once` pins the count. The `enabled` branches were correctly left un-collapsed.
- [x] ~~**`REG | diff-evidence | src/_core/infrastructure/notification/error_notifier.py:85 | notification-client-unreachable-in-safe-send`**~~ **FIXED** — option A taken: `notification_router` is its own `providers.Selector` (`core_container.py:494-504`) gated on `notification_warning_threshold is not None`, `disabled=providers.Object(None)`. Probe confirms `notification_router() -> None` and `_router -> None`, so the `else self._client` branch at `:85` is live again. Two-mode contract and docstring correctly kept as-is.
- [ ] **`GOV | rule-source | docs/operations/error-notifications.md | runbook-not-synced-with-channel-routing`** — `[OPEN][MEDIUM][GOV]`
  **basis:** rule-source — `AGENTS.md` § Drift Management
  **Impact:** carried forward, re-verified still open. The `:214` sub-point from round 1 is now **obsolete** — your HIGH fix restored `once per process, at the first dispatch attempt`. The rest stand: the env-var table (`:151-155`) omits all three new variables; the gating row (`:49`) still says everything below `NOTIFICATION_SEVERITY_THRESHOLD` is gated; the worker troubleshooting row (`:348`) now has a second remedy; the Related section (`:362-364`) still calls #286 "a follow-up". `docs/ai/shared/project-dna.md:624` likewise still ends "Channel routing is a follow-up (#286)."
  **Recommended fix:** maintainer-absorbed via `/sync-guidelines` after merge. **Not an ask for @siddhanttiwari19.**
- [ ] **`GOV | rule-source | PR #313 body § Governor Footer | footer-fails-ci-validate`** — `[OPEN][MEDIUM][GOV]`
  **basis:** rule-source — ADR 047 §D2 / `.github/workflows/governor-footer-lint.yml`
  **Impact:** CI `Validate` is red, and it is a real linter failure — not the fork-token comment noise the workflow already tolerates via `continue-on-error` (`:64-74`). The gate is the `Fail job` step (`:84-86`). Reproduced locally against the live body:

  ```text
  Governor Footer violations (2):
  body.md:37: expected '- <field>: <value>' (single space after '-' and ':')
    'links: https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/313, n/a'
  body.md:27: missing required field 'links'
  ```

  Two causes. The `links` line lost its leading `- `. **And the shape itself is invalid regardless** — `tools/check_governor_footer.py:186-195` accepts `n/a` only as the *entire* value; in a comma-separated list every entry must match `URL_RE`, so `<url>, n/a` still fails once the dash is restored (verified: `links entry must be an http(s) URL or 'n/a', got 'n/a'`). You were following `.github/pull_request_template.md:48`, which prescribes exactly that shape. **The template is wrong, so this is not your error** — tracked separately in #314.
  **Recommended fix:** maintainer edits the PR body; the workflow triggers on `edited`, so `Validate` re-runs with no push from you. **Not an ask for @siddhanttiwari19.**

### Coverage

- `[OK][REG]` both round-1 findings independently re-verified by probe against `69b95dd`, not accepted from the reply comments
- `[OK][CORR]` enabled path probed with env vars set before import — the container-capture gap documented in `test_optional_infra.py`. The selector-rereads-vs-kwarg-captures split matches every other provider in this container and cannot diverge in production (env read once at boot)
- `[OK][ARCH]` architecture-review-checklist §1 — no Domain → Infrastructure import; `NotificationRouter` sits in `infrastructure/notification/` and types against `BaseNotificationProtocol` from `_core/domain/protocols/`
- `[OK][ARCH]` §4 — `DeclarativeContainer` preserved; new providers use `Selector` + `Singleton` per ADR 042; `_build_notification_router` keeps the lazy-import convention
- `[OK][ARCH]` §5 — 25 new tests across router unit, notifier integration, container selectors, config, optional-infra; both round-1 fixes carry pinning regression tests
- `[OK][SEC]` security-checklist §13 — `_safe_send` still logs `exc_type` only; the two new webhook URLs are never logged; `_env` examples carry placeholders, not live URLs
- `[SKIP][CONTRACT]` no API schema, OpenAPI, status code, task payload, or migration touched

### Questions (not findings — no action required to merge)

1. `NOTIFICATION_WARNING_THRESHOLD` set with **no provider at all** is accepted at boot, which widens the gate to 4xx and spawns background tasks that only log `notification_suppressed`. Harmless, but a boot-time guard (threshold requires provider) would match the existing partial-config-group pattern. Carried from round 1 Q2.
2. Both `test_enabled_via_explicit_override` cases still set `slack_webhook_url=None` alongside a per-tier URL — a state boot validation rejects. Harmless, but it pins a configuration that cannot exist. Carried from round 1 Q3.
3. On the enabled path the base `notification_client` still builds a third live adapter that never posts (probe: `@ .../BASE`). Negligible, and option A was the right call — noting it only because it is what makes the `security-checklist.md:419` "sole production construction site" grep stale. Maintainer-side drift, already tracked.

### Drift Candidates (maintainer-side, recorded for `/sync-guidelines`)

| target | reason |
|---|---|
| `.github/pull_request_template.md:48` | prescribes a `links` shape the linter rejects — tracked in #314 |
| `docs/operations/error-notifications.md` | ledger item above |
| `docs/ai/shared/project-dna.md:624` | still calls channel routing a follow-up |
| `.claude/rules/{project-status,commands,project-overview,architecture-conventions}.md` | #286 under "Not Yet Implemented"; Error Notification sections lack the three new variables |
| `docs/ai/shared/security-checklist.md` §13 | `:419` "sole production construction site" claim; `:418`/`:427` greps miss `NOTIFICATION_{CRITICAL,WARNING}_WEBHOOK_URL` |
| `README.md:113`, `docs/reference.md:295` | feature line and checkbox omit channel routing |
| `_env/quickstart.env.example:120-123` | carries a `NOTIFICATION_*` block; only `local.env.example` was updated |

### Completion State

- summary-ledger: **unresolved** — 2 `OPEN` keys, both maintainer-absorbed
- 2 round-1 inline threads resolved; independently re-verified above
- merge blocked on the inline `HIGH` + CI `Validate`

### Sync Required

`true` — closed by `/sync-guidelines` after merge, on the maintainer side.

src/_core/infrastructure/di/core_container.py
Mr-DooSun: **FIXED** — verified independently at head `69b95dd`, not taken from the reply.

Shared `_noop_notification_client` Singleton at `core_container.py:448`, referenced from all three Selectors' `disabled` branches (`:458`, `:476`, `:487`). Same probe as the original comment, resolving `error_notifier()` exactly once with every `NOTIFICATION_*` setting unset:

```text
main   @ 1e301cf → notification_client_disabled ×1
pr/313 @ 69b95dd → notification_client_disabled ×1   (was ×3 at 0ce9b18)
```

`test_noop_notification_client_disabled_warning_emitted_once` pins the count, so it cannot silently return. The `enabled` branches were correctly left un-collapsed. This also restores the contract `docs/operations/error-notifications.md:214` documents.

src/_core/infrastructure/notification/error_notifier.py
Mr-DooSun: **FIXED** — verified independently at head `69b95dd`, not taken from the reply. Option A was the right pick.

`notification_router` is now its own `providers.Selector` (`core_container.py:494-504`) gated by `_notification_routing_selector` on `notification_warning_threshold is not None`, with `disabled=providers.Object(None)`. Probe on the default path:

```text
notification_router()   -> None
error_notifier._router  -> None
_effective_min_threshold() -> 500
```

So the `else self._client` branch on this line is live again in production, and the two-mode contract plus its docstring stay accurate — correct call to leave both as they were.

Confirmed the enabled side works too, probed the way production reads config (env vars set before import, since container kwargs capture `settings.*` at class-import time — the gap you documented in `test_optional_infra.py`):

```text
router.warning_threshold -> 400
resolve(404) -> SlackNotificationAdapter @ .../MONITORING
resolve(500) -> SlackNotificationAdapter @ .../ALERTS
resolve(399) -> None
```

One residual, no action needed: on the enabled path the base `notification_client` still builds a third live adapter that never posts (`@ .../BASE`). Negligible, and it is the maintainer-side `security-checklist.md:419` grep that needs updating, not this wiring.

src/_core/infrastructure/notification/error_notifier.py
siddhanttiwari19: Fixed — tier-scoped the cooldown key exactly as recommended.

Change: added _cooldown_key() — returns the bare error_code when self._router is None (byte-identical to #17's default path), and "{tier}:{error_code}" when routing is active. _should_notify now keys _last_notified_at off this instead of the bare code.

Test: inverted test_cooldown_is_shared_across_tiers_per_error_code → test_cooldown_is_scoped_per_tier (asserts both tiers now receive), and added test_cooldown_still_suppresses_within_same_tier as the within-tier guard.

Verified locally:

uv run pytest tests/unit/_core/infrastructure/notification/ tests/unit/_core/test_config.py tests/unit/_core/infrastructure/di/test_core_container_selectors.py tests/integration/_core/infrastructure/test_optional_infra.py -v

→ 197 passed

ruff check / ruff format --check → clean

Full suite also matches your round-2 numbers: 1545 passed, 32 skipped (same 10 DynamoDB errors, environmental / not from this change).

Pushed to feat/error-notification-channel-routing. Leaving the runbook drift and Governor Footer template issue on your side per your notes above.

Mr-DooSun:
Thank you for this — and for the patience across three rounds.

A few things worth naming specifically, because they were good calls rather than lucky ones:

- You took **option A** on the router gating when I offered two, which kept `ErrorNotifier`'s two-mode contract meaningful instead of hollowing it out.
- You **inverted** `test_cooldown_is_shared_across_tiers_per_error_code` into a real assertion rather than deleting it, and added `test_cooldown_still_suppresses_within_same_tier` alongside as the guard.
- You left the `enabled` Selector branches un-collapsed when I asked you not to touch them.
- You documented the container-kwarg capture gap in `test_optional_infra.py` honestly instead of papering over it — that docstring is the reason I could probe the enabled path correctly.

It is also only fair to say that **two of those three rounds were on me, not on you.** Round 1 audited the cross-tier cooldown case and closed it as "latent, no reachable path" — that premise was incomplete (`BaseCustomException` takes both `status_code` and `error_code` per instance), which is why it came back as round 2's blocker. And round 1's own required rewiring — unconditional `Singleton` to gated `Selector` — invalidated documentation that round 1 did not flag. You turned around every actual ask quickly, twice within a day.

This is going in. Three follow-ups are tracked as issues rather than a fourth round on this branch:

- **#315** (`bug`) — per-tier webhook URLs are silently ignored when `NOTIFICATION_WARNING_THRESHOLD` is unset; alerts land on the base webhook with no boot error and no log line. This is the only behavioural one. **If you'd like to take it, it's yours** — boot validation plus one test, and a cleaner PR than a fourth round here would have been.
- **#316** (docs) — `AGENTS.md:272` and four `src/` prose sites still describe the pre-round-1 wiring. Maintainer side.
- **#317** (docs) — shared-reference sync: runbook, `project-dna`, `.claude/rules/*`, security checklist. Maintainer side. This one I kept deferring across all three rounds, which is exactly why it is a tracked issue now instead of another review note.

One housekeeping note, for transparency: I edited the Governor Footer in the PR description. You had already fixed the `links` line yourself — I only updated `rounds` and the `r-points-*` counts to match the real review history. The shape you originally followed came straight from our own PR template, which is wrong; that is #314, and it is on us, not you.

If the project is useful to you, a ⭐ would be appreciated. And you are very welcome back — #315 is open if you want it, and the issue list has plenty else.

Issue #336, opened 2026-08-02
Project-wide audit (main @ ccd0f8a): 39 findings across 14 clusters


Mr-DooSun:
Project-wide audit of `main` @ `ccd0f8a`, run as a tiered pass: a 9-agent reconnaissance sweep,
then four parallel deep lanes, then adversarial verification of every blocking/high finding through
four independent lenses.

**Read-only.** No file in the repository was changed to produce this.

## Numbers

| | |
|---|---|
Scope | `src/` 18,389 LOC · `tests/` 25,730 LOC · 856 `.py` files, **plus** `_docker/docker.Dockerfile`, `docker-compose.yml`, `pyproject.toml`, `.github/workflows/ci.yml`, `_env/*.example` |
Recon floor | **28 verified findings** |
Deep pass | 34 raw → **11 survive** (4 high · 6 medium · 1 low) |
Adversarial verification | 4 lenses × every blocking/high · **0 refuted · 6 corrected · 2 downgraded** |
Pre-closed false positives | **~20**, each with its citation — see below |
Clusters | **14**, one issue each |

## The two findings worth reading first

**A subsystem shipped 2026-07-28 is inert on the default broker** (324). On
`BROKER_TYPE=inmemory` the server executes tasks through a broker with **zero middleware**, so
#310's worker alerts, retries, error logging and structlog context are all dead. `inmemory` is the
shipped default *and* passes stg/prod validation (`config.py:719-722`) — this was initially scoped
as dev-only and the refutation pass corrected it. In the same process, the one task the server can
dispatch dies on an unwired DI marker: an authenticated user uploads a large document, gets a 2xx,
and the row persists with `chunk_count=0` **forever**, permanently invisible to `/v1/docs/query`.

**A 5xx leaves no diagnostic** (323). `custom_exception_handler` makes no logger call, and
`Database.session()` puts the original error into `details` **only when `settings.is_dev`**. In
stg/prod the wrapped driver error exists in *none* of: the response, `details`, any log record, or
the Slack alert. One `http_request` WARNING survives, carrying no exception identity. **15 of 17**
5xx construction sites log nothing before raising — the two that do are the house pattern.

## Clusters

Ordered by the sequence they should be fixed in.

| Cluster | Issue | Contents | Tier | Depends on |
|---|---|---|---|---|
[...]

PR #339, opened 2026-08-02
fix(core): exception translation — log 5xx, gate the LLM mapper, curate provider errors

### LINKED ISSUE #323: Audit cluster F: exception translation — a 5xx leaves no diagnostic, and raw provider errors reach the response body
> Part of the project-wide audit at `main` @ `ccd0f8a`. Parent: #336
>
> Findings marked **[audit]** were produced by the deep pass and carry a probe with recorded
> output. Findings marked **[recon]** come from the reconnaissance sweep — verified, but re-run
> the probe before fixing since they were measured slightly earlier in the same commit.

## Why this is one cluster

One reviewer should hold the whole `raw exception → HTTP status → log → alert` path. F1 and the
`error_mapper` recon item are two halves of the same function; F5 is the same shape at the other
end of the chain.

**Land this early.** F1 degrades the diagnosability of every other finding in the audit — while it
is open, any production failure you go looking for may have left no trace.

---

### F1 — 5xx `BaseCustomException` produces no exception-level log record; the wrapped driver/provider error exists nowhere in stg/prod  [high][verified]

- **Location** `src/_core/exceptions/exception_handlers.py:72-89` (`custom_exception_handler`) and `:94-113` (mapped-LLM branch of `generic_exception_handler`); origin at `src/_core/infrastructure/persistence/rdb/database.py:207-217`
- **What** `custom_exception_handler` goes `jsonable_encoder` → `_dispatch_error_notification` → `JSONResponse` with no logger call. `generic_exception_handler` logs `unhandled_exception` **only** on the unmapped path (`:115`) — the `try_map_llm_error` branch returns at `:113` first. Meanwhile `Database.session()` puts `str(e)` into `details` **only when `settings.is_dev`**. In stg/prod the original error is in none of: the response body, `details`, any log record, or the Slack alert (which receives the curated `"500 [DB_INTERNAL_ERROR]: Internal database error"`). Exactly one record survives, from `RequestLogMiddleware` (`request_log_middleware.py:73-82`): an `http_request` WARNING with status/method/path/duration/request_id and **no exception identity**.
- **Why it matters** 17 distinct 5xx `BaseCustomException` construction sites exist across `_core` infra and three domains; **15 of them log nothing** before raising (the two exceptions — `dynamodb_client.py:75-89` and `vectors/s3/client.py:80-89` — prove the house pattern is to log). Deadlock, connection reset, `DataError`, S3 failure, LLM provider failure all produce a 500/502 with no diagnostic. The mapped-LLM branch compounds the floor's `error_mapper` substring problem: a plain socket `TimeoutError` whose message contains "rate limit" is returned as a 429 `LLM_RATE_LIMITED` and the misclassification leaves no trace, so operators cannot discover the mapper fired.
- **Probe**

```bash
cat > /tmp/p_f1.py <<'PY'
from fastapi import FastAPI
from fastapi.testclient import TestClient
from structlog.testing import capture_logs
from src._core.config import settings
from src._core.exceptions.base_exception import BaseCustomException
from src._core.exceptions.exception_handlers import custom_exception_handler, generic_exception_handler
from src._core.infrastructure.logging.request_log_middleware import RequestLogMiddleware
from src._core.infrastructure.di.core_container import CoreContainer
from src._core.infrastructure.persistence.rdb.database import Database
print("is_dev =", settings.is_dev, "| env =", settings.env)
cc = CoreContainer()
db = Database(database_engine="sqlite", database_user="", database_password="",
              database_host="", database_port=0, database_name=":memory:", config=cc.db_config())
app = FastAPI(); app.add_middleware(RequestLogMiddleware)
app.add_exception_handler(BaseCustomException, custom_exception_handler)
app.add_exception_handler(Exception, generic_exception_handler)
@app.get("/boom")
async def boom():
    async with db.session():
        raise RuntimeError("FATAL: deadlock detected on user_pkey")
with capture_logs() as logs:
    r = TestClient(app, raise_server_exceptions=False).get("/boom")
print("status:", r.status_code, r.text)
print("LOG RECORDS (full stack incl. RequestLogMiddleware):")
for rec in logs: print("  ", {k: v for k, v in rec.items() if k != "exc_info"})
print("any record containing driver text 'deadlock':", any("deadlock" in str(x) for x in logs))
PY
PYTHONPATH=. ENV=prod ADMIN_STORAGE_SECRET=a-real-admin-storage-secret-value-x \
 DATABASE_ENGINE=sqlite DATABASE_USER=app DATABASE_PASSWORD='db-s3cure-p@ss' \
 DATABASE_HOST=db.internal DATABASE_PORT=5432 DATABASE_NAME=':memory:' \
 JWT_SECRET_KEY=a-real-jwt-secret-key-with-enough-length \
 ADMIN_JWT_SECRET_KEY=a-distinct-admin-jwt-secret-key-len \
 TASK_NAME_PREFIX=myapp BROKER_TYPE=rabbitmq RABBITMQ_URL=amqp://g:g@h:5672/ \
 ALLOW_ORIGINS='["https://app.example.com"]' uv run python /tmp/p_f1.py
```

```
is_dev = False | env = prod
status: 500 {"success":false,"message":"Internal database error","errorCode":"DB_INTERNAL_ERROR","errorDetails":null}
LOG RECORDS (full stack incl. RequestLogMiddleware):
   {'duration_ms': 2.72, 'status_code': 500, 'event': 'http_request', 'log_level': 'warning'}
any record containing driver text 'deadlock': False
```

- **Not a documented decision** ADR 017 (`docs/history/017-exception-handling-strategy.md:96-103`) documents only *response* trace suppression — "Environment-aware trace in generic handler only … Only unexpected exceptions need trace information" — and its rationale rests on custom exceptions having "predictable formats", which a `DatabaseException` wrapping an arbitrary driver error does not. `docs/ai/shared/security-checklist.md:181-185` ("Stack Trace Exposure") governs response bodies. `AGENTS.md:276` and `tests/unit/_core/exceptions/test_exception_handlers_notification.py` pin *alerting*, not logging. Nothing in ADR 017, ADR 024, or project-dna §0/§8 says a 5xx should go unlogged, and the two sibling AWS clients already log.
- **Fix** `exception_handlers.py:72-89`:

```python
  content = jsonable_encoder(ErrorResponse(...))
+ if exc.status_code >= 500:
+     _logger.exception("custom_exception", exc_info=exc,
+                       error_code=exc.error_code, status_code=exc.status_code)
  _dispatch_error_notification(...)
```

Same in the mapped-LLM branch (`:94-113`), logging the *original* `exc` plus `mapped_error_code=mapped.error_code`.

- **Do not change** Do not widen the *response* — `errorDetails: null` in prod and the `settings.is_dev` gate on `details` are checklist-mandated (security-checklist:181-185). Do not move the log into `_dispatch_error_notification`; that helper is deliberately notifier-only and must keep swallowing everything. Do not log 4xx `BaseCustomException`s (404s from `select_data_by_id` are routine) — gate on `>= 500`. `raise ... from e` at `database.py:212` is **not** the fix: `__context__` is already set, so the traceback is already attached; only the rendering is missing.
- **Size** s · **Cluster** F

---

### F5 — `ObjectStorage` interpolates the raw boto3 `ClientError` into a 500 response body; IAM ARN, AWS account id, bucket and key would leak  [medium][verified]

- **Location** `src/_core/infrastructure/storage/object_storage.py:54, 70, 81, 94, 109, 124`
- **What** **All six** `except ClientError as e:` blocks build `BaseCustomException(status_code=500, message=f"Storage … failed: {e}")`. Two of the six have a curated *pre*-branch (`download_file:65-68` returns a 404 for `NoSuchKey`; `file_exists:91-92` returns False for `"404"`) but both still fall through to the raw-interpolated 500. `ErrorResponse.message` is serialized unconditionally by `custom_exception_handler` — unlike `details`, it has no `is_dev` gate. No `error_code` is passed, so every storage failure surfaces as generic `CUSTOM_ERROR`, and the same raw string is what `ErrorNotifier` posts to Slack/Discord. The two sibling AWS clients do the opposite (`vectors/s3/client.py:71-89`, `dynamodb_client.py:76-93`): extract `Error.Code`, log the raw provider message via structlog, return `"… operation failed [{error_code}]"` with a real error code and `from e`.
- **Why it matters** A boto3 `AccessDenied` message names the calling IAM principal ARN (embedding the 12-digit account id), the bucket, and the S3 key — the first two facts needed for a cross-account or role-confusion probe. **Downgraded from high because nothing reaches it at HEAD**: `rg -l 'ObjectStorage|object_storage' src/ examples/ tests/` returns only the DI wiring, the two storage modules, an unrelated substring in `dynamodb_client.py`, and `tests/integration/_core/test_minimal_install.py`; `rg 'UploadFile' src/ examples/` returns nothing. This is a latent hardening defect in provisioned-but-unconsumed infrastructure that becomes a live disclosure the moment an adopter wires the first upload route — which is exactly what a blueprint invites.
- **Probe**

```bash
uv run --extra aws python -c '<settings.env forced to prod; ObjectStorage(storage_client=<stub whose
client() raises a botocore AccessDenied ClientError with a realistic IAM message>,
bucket_name="prod-uploads"); await upload_file(b"x", key="tenant/42/passport.pdf");
feed the raised exception to custom_exception_handler inside capture_logs()>'
```

```
PROD(is_dev=False) status: 500
PROD response body: {"success":false,"message":"Storage upload failed: An error occurred (AccessDenied) when calling the PutObject operation: User: arn:aws:iam::123456789012:user/app-svc is not authorized to perform: s3:PutObject on resource: arn:aws:s3:::prod-uploads/tenant/42/passport.pdf","errorCode":"CUSTOM_ERROR","errorDetails":null}
log records: []
```

Reachability confirmed separately:

```
$ rg -l 'ObjectStorage|object_storage' src/ examples/ tests/
tests/integration/_core/test_minimal_install.py
src/_core/infrastructure/storage/object_storage_client.py
src/_core/infrastructure/di/core_container.py
src/_core/infrastructure/storage/object_storage.py
src/_core/infrastructure/persistence/nosql/dynamodb/dynamodb_client.py
$ rg -n 'UploadFile' src/ examples/     # (no output)
```

- **Not a documented decision** `docs/ai/shared/security-checklist.md:184` — "[Always][HIGH] Error responses do not expose internal implementation details" — and `:187-189` "Sensitive Data in Error Messages"; `:409` requires the same for the LLM analogue. ADR 017 states "Production: returns only 'Internal server error' with no internal details". No ADR covers storage error handling (ADR 009 mentions only file layout); `grep -rn 'Storage upload failed' docs/` returns nothing. The two sibling AWS clients in this repo already implement the curated pattern, so this is an outlier, not a convention. The **non-goal** citation is about the *feature*, not the pattern: project-dna §0 lists file upload under "ship on demonstrated demand" and `.claude/rules/project-status.md` lists "File Upload (UploadFile)" as Not Yet Implemented — that is what caps severity, not what excuses the message.
- **Fix** Each of the six handlers, mirroring `vectors/s3/client.py:71-89`:

```python
- except ClientError as e:
-     raise BaseCustomException(status_code=500, message=f"Storage upload failed: {e}")
+ except ClientError as e:
+     error_code = e.response.get("Error", {}).get("Code", "Unknown")
+     _logger.error("storage_operation_failed", operation="upload", error_code=error_code,
+                   error_message=e.response.get("Error", {}).get("Message", str(e)))
+     raise BaseCustomException(status_code=500,
+                               message=f"Storage operation failed [{error_code}]",
+                               error_code="STORAGE_OPERATION_FAILED") from e
```

- **Do not change** Do not gate the message on `settings.is_dev` — that repeats the `database.py` pattern whose failure mode is F1 (the detail then exists nowhere in prod). Log the raw message **and** curate the response; both, not either. Keep the `ClientError` ImportError shim at `:9-22` exactly as is (documented no-aws-extra boot path). Keep `status_code=500` — a storage failure is genuinely server-side. Preserve the two curated pre-branches (404 for `NoSuchKey`, `return False` for head-object 404).
- **Size** s · **Cluster** F

---

### [recon] `error_mapper._classify` substring-matches every unhandled exception

- **Location** `src/_core/infrastructure/llm/error_mapper.py:60-88`, reached from `exception_handlers.py:100-102`
- **What** `_classify()` lowercases `str(exc)` and substring-matches it against **every** unhandled
  application exception, then returns early — skipping both `_logger.exception` and (being 4xx) the
  notification.
- **Probe output (recon)** `ValueError("...row limit exceeded")` → 429; `KeyError("unauthorized")` → 401;
  `RuntimeError("S3 bucket model-artifacts not found")` → 404.
- **Fix** gate the mapper on the exception's originating provider module rather than its message text,
  and move `_logger.exception` above the mapping branch. **These two are one defect with two halves** —
  narrowing the match alone leaves genuine 500s unlogged whenever it still fires.

### [recon] `http_client` except ordering makes the 504 path unreachable

- **Location** `src/_core/infrastructure/http/http_client.py:72-78`
- **What** `except aiohttp.ClientError` precedes `except TimeoutError`. In aiohttp 3.13.5 the
  connect/sock_read timeout classes subclass **both**, so `ExternalServiceTimeoutException` (504) is
  unreachable for 2 of the 3 configured timeouts.
- **Fix** reorder the handlers. Verify against the installed aiohttp version's MRO, not the docs.

## Sequencing

No dependency. Land early — F1 is a prerequisite for diagnosing anything else in production.

## Governor tier

None. The security-checklist §13 wording edit that F5 implies routes to cluster J.

Mr-DooSun:
Resolves #323 (audit cluster F). Parent: #336

Cluster F is the `raw exception → HTTP status → log → alert` path. All four issue
findings are fixed, plus one adjacent leak found while fixing them.

## What changed

**1. The LLM error mapper classified non-provider exceptions** (`error_mapper.py`)

`generic_exception_handler` hands `try_map_llm_error` *every* unhandled
application exception, and `_classify` matched on message substrings with no
check on origin. Re-probed at `d5c2a1d`:

```
KeyError("unauthorized")                            -> 401 LLM_AUTH_FAILED
RuntimeError("S3 bucket model-artifacts not found") -> 404 LLM_MODEL_NOT_FOUND
TimeoutError("... after rate limit backoff")        -> 429 LLM_RATE_LIMITED
```

Each is a genuine 5xx returned as a 4xx — and *because* it is a 4xx it also
skipped the exception log and the error notification, so the misclassification
left no trace anywhere. Now gated on `type(exc).__module__` matching a provider
SDK prefix.

Bedrock decided the shape. Its exceptions are generated at runtime by botocore,
so they are not importable and cannot be matched by identity — but they do carry
`__module__ == "botocore.errorfactory"`. Prefix matching alone suffices, which
keeps this module free of a botocore import. A test builds a real
`bedrock-runtime` `ThrottlingException` to pin that.

The module had **no tests**. Adds 25, covering both directions.

**2. A 5xx left no diagnostic** (`exception_handlers.py`) — F1

`custom_exception_handler` made no logger call; in `generic_exception_handler`
the `try_map_llm_error` branch returned before `_logger.exception`. For a curated
5xx the wrapped cause was in *none* of: the response body (curated by design),
`details` (only when `settings.is_dev`), any log record, or the notification
(same curated text). In stg/prod the sole survivor was `RequestLogMiddleware`'s
`http_request` line — status and duration, no exception identity.

Gated at `>= 500` as the issue specifies: curated 4xx are routine and logging
them at error level would bury the signal this adds. Tests pin that the response
body is unchanged **in both directions**, including the ADR 017 dev-only trace in
`errorDetails` — an absolute "no raw text reaches the client" assertion would be
wrong, and my first draft of that test was.

**3. Raw provider text reached response bodies** — F5 + one adjacent finding

`object_storage.py` — all six `except ClientError` blocks. A boto3 `AccessDenied`
message names the calling IAM principal ARN (embedding the 12-digit account id),
the bucket and the key. Routed through one `_storage_failure` helper following
the `dynamodb_client.py` / `vectors/s3/client.py` house pattern, plus a real
`STORAGE_OPERATION_FAILED` code (every storage failure was generic
`CUSTOM_ERROR`) and `from e`. `Error.Code` extraction also moves off
`e.response["Error"]["Code"]`, which raises `KeyError` *inside an except block*
for any error whose response lacks the key.

`http_client.py:73` — **not in the issue.** Same shape: `f"External service
error: {e}"`. Verified on aiohttp 3.13.5 that the message carries request detail:

```
403, message='Forbidden', url='https://hooks.slack.com/services/T/B/SECRET'
Cannot connect to host internal-svc.prod:8443 ssl:default [None]
```

`security-checklist.md` §13 records that property for the log stream; it applies
just as much to a response body, and for a notification webhook the URL *is* the
credential. The client now gets the failure class and upstream status.

The log side needed care: **both webhook adapters POST through this client**, and
`error_notifier.py` logs `exc_type` only precisely so the webhook URL never
reaches logs. Logging `str(exc)` here would have re-introduced the leak #17
deliberately avoided. So the log gets the origin only — `scheme://host[:port]`,
path and query dropped, since a Slack token lives in the path and the host is
not secret.

**4. The 504 branch was unreachable** (`http_client.py`) — except ordering

`except aiohttp.ClientError` preceded `except TimeoutError`. All three aiohttp
timeout classes subclass **both**, so every timed-out upstream was reported as a
502. The issue framed this as partial; the probe showed it is total:

```
ServerTimeoutError       ClientError / TimeoutError
SocketTimeoutError       ClientError / TimeoutError
ConnectionTimeoutError   ClientError / TimeoutError
```

## Verification

- `1637 passed, 42 skipped` (from `1585`; +52 tests). Zero failures, zero errors.
- The new tests were checked against the **unpatched** modules: 10 fail there,
  15 pass after. They catch the findings rather than describing the fix.
- `pre-commit run` clean; `mypy` (manual stage) clean — it caught a `str` /
  `Any | None` rebind in `_safe_origin` that I fixed.
- `check_governor_footer.py --require-governor-footer` run against the changed
  file list: `0 violations`, non-governor. Run, not assumed — asserting this
  without running it is what broke CI on #338.
- The pre-existing `format_exc_info` structlog warning in the docs e2e test was
  confirmed present *before* these changes, so it is not from this PR.

## Deliberately not changed

- The response is not widened — `errorDetails: null` in prod and the
  `settings.is_dev` gate stay as security-checklist:181-185 mandates.
- The log did not move into `_dispatch_error_notification`; that helper stays
  notifier-only and keeps swallowing everything.
- `raise ... from e` at `database.py:212` is not the fix — `__context__` is
  already set, so the traceback was already attached; only rendering was missing.
- `file_exists` still tests `"404"` and not `"NoSuchKey"`: HEAD has no response
  body, so botocore falls back to the HTTP status as the code. Only the
  extraction became KeyError-safe; the semantics are untouched.

### DIFF
--- a/src/_core/exceptions/exception_handlers.py
+++ b/src/_core/exceptions/exception_handlers.py
@@ -13,6 +13,11 @@ from src._core.exceptions.base_exception import BaseCustomException
 
 _logger = structlog.stdlib.get_logger("src._core.exceptions")
 
+# Status at or above which a curated exception is also an operational event worth
+# an exception-level log record. Named rather than inlined so the worker path can
+# share the same boundary (see taskiq_middleware._synthetic_status_code).
+_SERVER_ERROR_FLOOR = 500
+
 
 def _dispatch_error_notification(
     request: Request, *, status_code: int, error_code: str, message: str
@@ -81,6 +86,19 @@ async def custom_exception_handler(
             error_details=exc.details,
         )
     )
+    # A curated 5xx still means something broke. Without this record the wrapped
+    # cause exists nowhere in stg/prod: the response is curated by design, the
+    # notification receives that same curated text, and ``details`` carries the
+    # original only when ``settings.is_dev``. Curated 4xx are normal traffic and
+    # stay unlogged — logging them at error level would bury this signal.
+    if exc.status_code >= _SERVER_ERROR_FLOOR:
+        _logger.exception(
+            "custom_exception",
+            exc_info=exc,
+            exception_type=type(exc).__name__,
+            error_code=exc.error_code,
+            status_code=exc.status_code,
+        )
     _dispatch_error_notification(
         request,
         status_code=exc.status_code,
@@ -97,6 +115,19 @@ async def generic_exception_handler(request: Request, exc: Exception) -> JSONRes
 
     mapped = try_map_llm_error(exc)
     if mapped is not None:
+        # Log before the early return. This branch used to bypass
+        # ``_logger.exception`` below, so a provider error — or, before the
+        # mapper was gated on provider modules, a *misclassified* ordinary
+        # exception — produced a 4xx with no trace anywhere. The original ``exc``
+        # is what matters here; ``mapped`` is a translation of it.
+        _logger.exception(
+            "mapped_provider_exception",
+            exc_info=exc,
+            exception_type=type(exc).__name__,
+            exception_module=type(exc).__module__,
+            mapped_error_code=mapped.error_code,
+            mapped_status_code=mapped.status_code,
+        )
         content = jsonable_encoder(
             ErrorResponse(
                 message=mapped.message,
--- a/src/_core/infrastructure/http/http_client.py
+++ b/src/_core/infrastructure/http/http_client.py
@@ -3,11 +3,72 @@ from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 
 import aiohttp
+import structlog
 
 from src._core.infrastructure.http.exceptions import (
     ExternalServiceException,
     ExternalServiceTimeoutException,
 )
+
+_logger = structlog.stdlib.get_logger(__name__)
+
+_DEFAULT_PORTS = frozenset({80, 443})
+
+
+def _safe_origin(exc: aiohttp.ClientError) -> str | None:
+    """``scheme://host[:port]`` for the failed request — path and query dropped.
+
+    The path is the secret half of a webhook URL (a Slack token lives in
+    ``/services/T/B/X``); the origin is not. Keeping the origin makes the log
+    actionable without re-introducing the leak that ``error_notifier.py`` avoids
+    deliberately: both webhook adapters POST through this client, so anything
+    logged here can be a credential.
+
+    Returns ``None`` for the errors that carry no request context at all
+    (``ClientPayloadError`` has neither ``request_info`` nor ``host``).
+    """
+    request_info = getattr(exc, "request_info", None)
+    url = getattr(request_info, "url", None)
+    if url is not None:
+        suffix = f":{url.port}" if url.port not in _DEFAULT_PORTS else ""
+        return f"{url.scheme}://{url.host}{suffix}"
+
+    # ClientConnectorError never got a response, so it exposes host/port instead.
+    host = getattr(exc, "host", None)
+    if host is not None:
+        port = getattr(exc, "port", None)
+        if port is None or port in _DEFAULT_PORTS:
+            return str(host)
+        return f"{host}:{port}"
+
+    return None
+
+
+def _curate_client_error(exc: aiohttp.ClientError) -> ExternalServiceException:
+    """Build the 502 for an aiohttp failure without copying its message.
+
+    ``f"External service error: {exc}"`` published request detail to the caller:
+    ``ClientResponseError`` renders the **full URL** and ``ClientConnectorError``
+    the internal ``host:port``. Verified on aiohttp 3.13.5::
+
+        403, message='Forbidden', url='https://hooks.slack.com/services/T/B/X'
+        Cannot connect to host internal-svc.prod:8443 ssl:default [None]
+
+    ``security-checklist.md`` §13 records that property for the log stream; it
+    applies just as much to a response body. The client now gets the failure
+    class and the upstream status; the origin goes to the log only.
+    """
+    status = getattr(exc, "status", None)
+    _logger.error(
+        "external_service_error",
+        exc_type=type(exc).__name__,
+        upstream_status=status,
+        origin=_safe_origin(exc),
+    )
+    detail = f"{type(exc).__name__}"
+    if status is not None:
+        detail = f"{detail} {status}"
+    return ExternalServiceException(message=f"External service error [{detail}]")
 
 
 def get_http_client_config(env: str):
@@ -70,10 +131,20 @@ class HttpClient:
         try:
             session = await self._ensure_session()
             yield session
+        # Order is load-bearing: on aiohttp 3.13.5 every timeout class
+        # (ServerTimeoutError, SocketTimeoutError, ConnectionTimeoutError)
+        # subclasses *both* TimeoutError and aiohttp.ClientError, so catching
+        # ClientError first made the 504 branch unreachable for all three — a
+        # timed-out upstream was reported as a 502.
+        except TimeoutError as e:
+            _logger.error(
+                "external_service_timeout",
+                exc_type=type(e).__name__,
+                origin=_safe_origin(e) if isinstance(e, aiohttp.ClientError) else None,
+            )
+            raise ExternalServiceTimeoutException() from e
         except aiohttp.ClientError as e:
-            raise ExternalServiceException(message=f"External service error: {e}")
-        except TimeoutError:
-            raise ExternalServiceTimeoutException()
+            raise _curate_client_error(e) from e
 
     async def dispose(self) -> None:
         if self._client_session and not self._client_session.closed:
--- a/src/_core/infrastructure/llm/error_mapper.py
+++ b/src/_core/infrastructure/llm/error_mapper.py
@@ -58,12 +58,51 @@ _NOT_FOUND_NAMES = frozenset(
 )
 
 
+_PROVIDER_MODULE_PREFIXES: tuple[str, ...] = (
+    "openai",
+    "anthropic",
+    "botocore",  # Bedrock — see _is_provider_exception
+    "boto3",
+    "pydantic_ai",
+    "google.api_core",
+    "google.genai",
+    "google.generativeai",
+)
+
+
+def _is_provider_exception(exc: Exception) -> bool:
+    """Did this exception come out of an LLM provider SDK?
+
+    This gate exists because the heuristics below match on message *text*, and
+    ``generic_exception_handler`` hands this function **every** unhandled
+    application exception. Ungated, that mapped ordinary failures to LLM 4xx:
+    ``KeyError("unauthorized")`` became a 401, ``RuntimeError("S3 bucket
+    model-artifacts not found")`` a 404. Both are genuine 5xx, and being 4xx
+    they also skipped the exception log and the error notification — so the
+    misclassification left no trace anywhere.
+
+    Bedrock is the case that shapes this. Its exceptions are generated at
+    runtime by botocore, so they are not importable and cannot be matched by
+    identity — but they do carry ``__module__ == "botocore.errorfactory"`` and
+    ``botocore.exceptions.ClientError`` in their MRO. Module prefix alone is
+    enough; the MRO is not consulted, which keeps this free of a botocore import.
+    """
+    module = type(exc).__module__ or ""
+    return any(
+        module == prefix or module.startswith(f"{prefix}.")
+        for prefix in _PROVIDER_MODULE_PREFIXES
+    )
+
+
 def _classify(exc: Exception) -> LLMException | None:
     """Classify an exception as a known LLM provider error.
 
     Returns a domain LLMException instance, or None if the exception is not
     a recognisable provider error (caller should treat it as a generic 500).
     """
+    if not _is_provider_exception(exc):
+        return None
+
     type_name = type(exc).__name__.lower().replace("_", "")
     error_str = str(exc).lower()
 
--- a/src/_core/infrastructure/llm/error_mapper.py
+++ b/src/_core/infrastructure/llm/error_mapper.py
@@ -61,12 +61,27 @@ _NOT_FOUND_NAMES = frozenset(
 _PROVIDER_MODULE_PREFIXES: tuple[str, ...] = (
     "openai",
     "anthropic",
-    "botocore",  # Bedrock — see _is_provider_exception
-    "boto3",
     "pydantic_ai",
     "google.api_core",
     "google.genai",
     "google.generativeai",
+)
+
+# Bedrock cannot be matched by module: botocore is shared by *every* AWS client,
+# so `botocore.*` would readmit exactly the misclassification this gate exists to
+# stop — a real S3 `GetObject` failure whose message contains "unauthorized"
+# became a 401 `LLM_AUTH_FAILED`, and a DynamoDB `Query` throttle became a 429.
+# `botocore.errorfactory` does not separate them either: `s3.exceptions.NoSuchKey`
+# and `bedrock-runtime`'s `ThrottlingException` are both generated into that same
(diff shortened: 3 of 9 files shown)


Mr-DooSun:
Cross review (codex, read-only) returned **FAIL** with two HIGH findings. Both were real and are fixed in `07c043c` and `37b8175`.

## 1. `botocore` as a module prefix reopened the bug for AWS  [HIGH][CORR]

The gate I added allowed all of `botocore`, and every AWS client raises `botocore.exceptions.ClientError`. Reproduced:

```
S3 GetObject + "unauthorized"     -> 401 LLM_AUTH_FAILED
DynamoDB Query + "Rate exceeded"  -> 429 LLM_RATE_LIMITED
```

That is the exact misclassification the gate exists to stop, readmitted for S3 and DynamoDB. My mistake — I reached for the module because Bedrock exceptions are not importable, without checking who else lives in that module.

The review proposed keying on `module == "botocore.errorfactory"`. **That still leaks**, and I checked against a live botocore session before applying it:

```
s3.exceptions.NoSuchKey.__module__                == botocore.errorfactory
bedrock.exceptions.ThrottlingException.__module__ == botocore.errorfactory
```

So the botocore branch keys on `operation_name` instead, which `ClientError.__init__` sets on the dynamically generated subclasses too. Only `InvokeModel`, `InvokeModelWithResponseStream`, `Converse`, `ConverseStream` count. `boto3` came off the prefix list as well. 9 new tests: 5 non-Bedrock AWS operations that must not map, 4 Bedrock operations that must.

## 2. `Error.Message` as a structlog kwarg  [HIGH][SEC]

Correct, and `security-checklist.md:194` is explicit: structlog kwargs carry no sensitive fields. The string names the IAM principal ARN, the account id, and — for S3 — the object key, which for a user-uploaded file is often personal data itself.

I did **not** take the proposed patch as written. It also changed `raise ... from e` to `from None`, and that contradicts the other half of this PR: the 5xx handler now logs with `exc_info` precisely because the wrapped cause has nowhere else to live. Suppressing the chain would re-open F1 while closing this.

The split applied instead: **kwargs stay clean, the chain keeps the text.** `error_code` and a new `provider_exception_type` are logged; the provider message reaches the log only through the rendered `__cause__`. Applied to all three AWS clients, not just `object_storage`, so the family stays consistent.

The tests assert that precisely — kwargs are clean, *and* `exc.__cause__` still holds the ARN. An absolute "the ARN is not in the log output" assertion would be false, and claiming it would be the same over-assertion mistake that cost three rounds on #321.

## Not changed

- `httpx` is deliberately absent from the provider prefixes. An `httpx.HTTPStatusError` for a 401 renders "Unauthorized" in its message, so adding it would re-create the bug for every non-LLM outbound call. A raw httpx error from an LLM path now becomes a logged, alerted 500 rather than a guessed 4xx — the intended failure mode.
- No `tenacity`/`backoff` in `src/`, so the retry-wrapper false-negative the review raised has no live path. Worth remembering if one is introduced.

Suite: `1648 passed, 42 skipped, 0 failed`. mypy manual-stage clean.

Issue #324, opened 2026-08-02
Audit cluster L: on the default in-memory broker the server runs tasks with zero middleware, and docs ingestion silently no-ops


Mr-DooSun:
> Part of the project-wide audit at `main` @ `ccd0f8a`. Parent: #336
>
> Findings marked **[audit]** were produced by the deep pass and carry a probe with recorded
> output. Findings marked **[recon]** come from the reconnaissance sweep — verified, but re-run
> the probe before fixing since they were measured slightly earlier in the same commit.

## Why this is one cluster

The same gap from two angles — middleware wiring and DI wiring — sharing
`src/_apps/server/bootstrap.py`. Fixing either alone leaves the other's failure silent.

**This is the audit's most surprising result.** `#310` worker error notifications shipped
2026-07-28 and are inert on the shipped-default broker, which also passes stg/prod validation.

---

### F2 — On `BROKER_TYPE=inmemory` the server executes tasks through a broker with zero middleware; #310 worker alerts, retries, error logging and structlog context are all inert  [high][verified]

- **Location** `src/_apps/worker/bootstrap.py:58-85`; `src/_apps/worker/broker.py:1-5`; `src/_apps/server/bootstrap.py`
- **What** `StructlogContextMiddleware`, `PermanentAwareSmartRetryMiddleware`, `TaskErrorLoggingMiddleware` and `TaskFailureNotificationMiddleware` are installed only by `_install_middleware` in the worker bootstrap, reached only via `bootstrap_app` ← `src/_apps/worker/app.py` — a module the server never imports. The server imports `src/_apps/worker/broker.py`, which builds its **own** module-level `CoreContainer()` and resolves `broker`, with nothing calling `bootstrap_app` on it. Under `inmemory`, that unmiddlewared broker is what executes tasks inline in the server process. Taskiq runs `on_error` over `reversed(broker.middlewares)`; an empty list means no hook fires, so `maybe_dispatch` is never called regardless of `NOTIFICATION_*` configuration.
[...]

PR #340, opened 2026-08-02
fix(admin): share the server's CoreContainer instead of building a second tree

### LINKED ISSUE #326: Audit cluster C: a server with the admin extra runs two CoreContainers, and the test override reaches only one
> Part of the project-wide audit at `main` @ `ccd0f8a`. Parent: #336
>
> Findings marked **[audit]** were produced by the deep pass and carry a probe with recorded
> output. Findings marked **[recon]** come from the reconnaissance sweep — verified, but re-run
> the probe before fixing since they were measured slightly earlier in the same commit.

## Why this is one cluster

Container topology, audit-logger DI and the e2e override are one file set and one PR. Cluster **L**
must assume C's final topology, so land C first.

---

### F7 — `override_database` reaches only the server `CoreContainer`; the admin tree, including the eagerly-bound audit-log repository, keeps the real `DATABASE_*` connection  [medium][verified]

- **Location** `src/_apps/server/testing.py:31-38,66-68`; `src/_apps/admin/bootstrap.py:49,69-72`; `src/_apps/admin/di/container.py:7-25`
- **What** `bootstrap_admin` builds a second independent `CoreContainer` tree via `create_admin_container()`, then at `:69` **eagerly** resolves `admin_container.core_container.database()` and hands the concrete instance to `AdminAuditLogRepository`, stored in module globals by `configure_audit_repository`. `override_database(app, test_db)` overrides `app.state.container.core_container().database` — the server tree only. The admin tree, its per-domain repositories, and the already-constructed audit repository are unreachable. Beyond `database`, every `CoreContainer` Singleton duplicates: `http_client`, `broker`, `taskiq_manager`, `embedding_client`, `llm_model`, `notification_client`, `error_notifier`. Only the `providers.Object(None)` disabled branches are shared.
- **Why it matters** (1) **Test split-brain**: any harness that swaps the database gets `/v1/*` on the swapped DB while `/admin/*` and every audit write hit the real configured `DATABASE_*` — in the SQLite default, `./quickstart.db` in the repo working directory. Latent today because `tests/e2e/` reaches admin services only via `app.state.container`, but the first e2e test driving a NiceGUI page or asserting on audit rows breaks silently. (2) **Production**: two `Database` objects mean two async engines and two QueuePools per process. `database_pool_size`/`database_max_overflow` are unset by default, so 2 × (5 + 10) = up to 30 connections per process against a 15-connection budget.
- **Probe**

```bash
PYTHONPATH=. ENV=local DATABASE_ENGINE=sqlite uv run python -c "
from src._apps.server.app import app
from src._apps.server.testing import override_database
from src._core.infrastructure.persistence.rdb.database import Database
from src._core.infrastructure.admin.audit.logger import get_audit_repository
cc = app.state.container.core_container()
test_db = Database(database_engine='sqlite', database_user='', database_password='',
                   database_host='', database_port=0, database_name=':memory:', config=cc.db_config())
override_database(app, test_db)
print('C) server tree db is test_db :', cc.database() is test_db)
print('D) admin  tree db is test_db :', app.state.admin_container.core_container.database() is test_db)
r = get_audit_repository()
print('E) audit repo db is test_db  :', r._database is test_db, '->', str(r._database.engine.url))
print('F) admin user_repo db is test:', app.state.admin_container.user_container.user_repository().database is test_db)
print('G) server user_repo db is test:', app.state.container.user_container.user_repository().database is test_db)
print('H) engines distinct          :', id(cc.database().engine), id(app.state.admin_container.core_container.database().engine))"
```

```
C) server tree db is test_db : True
D) admin  tree db is test_db : False
E) audit repo db is test_db  : False -> sqlite:///./quickstart.db
F) admin user_repo db is test: False
G) server user_repo db is test: True
H) engines distinct          : 4503915088 4503783136
```

- **Not a documented decision** `rg 'admin_container|two CoreContainer|separate CoreContainer|create_admin_container' docs/ AGENTS.md .claude/rules/` yields one hit — `docs/history/archive/026-nicegui-admin-dashboard.md:174`, a `_service_provider` snippet saying nothing about a second core tree. ADR 007 mentions `admin/` only as a directory in a tree listing and describes `create_server_container()` as the pattern. project-dna §8 and `.claude/rules/project-overview.md` § App Entrypoints describe admin as "mounted on server via `ui.run_with`" — one process — with no note of a duplicate container graph. Notably `create_worker_container(core_container)` already takes the core container as a parameter, so the injected-core pattern exists in-repo and admin simply does not use it.
- **Fix** Adopt the worker's pattern:

```python
- def create_admin_container() -> containers.DynamicContainer:
-     container.core_container = providers.Container(CoreContainer)
+ def create_admin_container(core_container) -> containers.DynamicContainer:
+     container.core_container = core_container
# src/_apps/admin/bootstrap.py:49
- admin_container = create_admin_container()
+ admin_container = create_admin_container(fastapi_app.state.container.core_container)
```

Additionally, stop freezing a concrete `Database` into module state — `configure_audit_repository` should take the **provider** and resolve per call, so overrides applied after bootstrap are honoured. If the shared-core change is deferred, at minimum make `override_database` iterate every `CoreContainer` tree hanging off `app.state`.

- **Do not change** Do not delete `create_admin_container()` — its domain auto-discovery loop is what populates `admin_container.<domain>_container` for `_discover_and_register_pages` (`bootstrap.py:150-152`), and NiceGUI page registration happens at bootstrap time, before routing. Keep the `DynamicContainer` and the discovery loop; share only the `core_container` provider. Do not collapse admin domain containers into the server's — `page_config._service_provider` deliberately holds a provider, not a resolved service, and that indirection must survive. Note the process actually holds **three** `CoreContainer` trees (see F2); land this before F2 so the inline-middleware fix targets the final topology.
- **Size** m · **Cluster** C

---

### [recon] The duplication is wider than the database

- **Location** `src/_apps/admin/di/container.py:7-25`
- **What** `create_admin_container()` builds its own `providers.Container(CoreContainer)`, so a server
  with the admin extra holds **two** CoreContainers: 2 sync + 2 async engines, 2 connection pools, 2
  `HttpClient`s, 2 notifier cooldown dicts (so the per-process cooldown is really per-*tree*).
- **Call chain** `server/bootstrap.py:46 → _mount_admin_if_available:176 → admin/bootstrap.py:49 →
  core_container.database():70`.

## Open decision

Share the core container using the worker's existing `create_worker_container(core_container)`
pattern, versus making `override_database` walk every tree on `app.state`. **Only the first fixes the
production pool doubling.** If admin genuinely wants its own pool budget, that should become a stated
decision with explicit `database_pool_size` values rather than an accident of construction.

## Sequencing

Land **before L**. The four service-locator globals in `_core/infrastructure/admin` and any test
touching the admin tree must move in the same wave or immediately after.

## Governor tier

None.

Mr-DooSun:
Resolves #326 (audit cluster C). Parent: #336

Cluster C is container topology. Land it before #324 (cluster L) so the inline-broker middleware fix targets the final shape.

## What was wrong

A server with the `admin` extra held **two** `CoreContainer` trees. Re-verified at `d5c2a1d`:

```
C) server tree db is test_db  : True
D) admin  tree db is test_db  : False
E) audit repo db is test_db   : False -> sqlite:///./quickstart.db
F) admin user_repo db is test : False
H) engines distinct           : True
```

Two consequences, one latent and one live:

- **Test split-brain.** `override_database` reached the server tree only, so a harness that swapped the database got `/v1/*` on the swapped one while `/admin/*` and every audit write kept the real `DATABASE_*`.
- **Production.** Two async engines and two QueuePools per process. With `database_pool_size`/`database_max_overflow` unset that is up to 2 × (5 + 10) = 30 connections against a 15-connection budget, plus two `HttpClient`s and two `ErrorNotifier` cooldown dicts — making the "per-process" notification cooldown really per-*tree*.

**One thing the issue did not record**, found while probing: every optional-infra disabled warning fired twice, including the `notification_client_disabled` line. #286 made three notification Selectors share one Noop Singleton specifically so that line appears exactly once, and `test_noop_notification_client_disabled_warning_emitted_once` pins it. The test passed only because it builds a single container; the shipped server violated the invariant.

## The issue's prescribed fix does not work

The issue said to adopt `create_worker_container`'s injected-core pattern: `create_admin_container(fastapi_app.state.container.core_container)`. I implemented it and it raises at boot:

```
AttributeError: 'DependenciesContainer' object has no attribute '__self__'
```

`providers.Container(DomainContainer, core_container=...)` **overrides class-level state** on the domain container. By the time admin mounts, the server has already applied that override; re-applying it with a core container that is already in use fails. Tested against the real boot, in order of preference:

```
fresh providers.Container(CoreContainer)  -> OK    <- the current, broken behaviour
the server's provider                    -> AttributeError
providers.Object(resolved instance)       -> AttributeError
the resolved instance                     -> AttributeError
CoreContainer() brand new                -> OK    <- also the broken behaviour
```

Only a *brand-new* tree is accepted as the second override — the very thing being removed. The worker pattern works because a worker process wires those classes exactly once.

So admin now **aliases the server's existing providers** rather than constructing its own. The issue's "do not collapse admin domain containers into the server's" note gave its reason as preserving the `page_config._service_provider` indirection; aliasing preserves that exactly, because the values stay providers and are never resolved. A test pins it.

## Also fixed

`bootstrap_admin` resolved `core_container.database()` and `configure_audit_repository` froze that instance into module globals, so even with a shared tree an override applied *after* bootstrap stayed invisible. `AdminAuditLogRepository` now accepts a `Database` **or** a zero-arg provider and resolves per access. The worker cleanup task and the five existing unit-test call sites pass a `Database` directly and are unchanged.

## After

```
D) admin  tree db is test_db  : True
E) audit repo db is test_db   : True -> sqlite:///:memory:
F) admin user_repo db is test : True
H) engines distinct           : False
CoreContainer trees           : 1  (was 2)
all 7 core Singletons shared  : True
```

## Verification

- `1597 passed, 42 skipped, 0 failed` — standalone and in the full suite.
- 11 of the 12 new tests fail against the previous implementation.
- `pre-commit run` clean. `check_governor_footer.py --require-governor-footer` run against the changed file list: `0 violations`, non-governor.
- mypy reports 7 errors in `src/_apps/admin/` — all **pre-existing**, verified identical on `origin/main` (only line numbers shift). Not introduced here and not fixed here.

## One test-writing note worth keeping

The override assertions are deliberately **one** test with an explicit `user_repository.reset()`. `user_repository` is a `Singleton`, and the autouse `_override_app_database` in `tests/e2e` resolves it first, so a split version passed in isolation and failed in the full suite for a reason unrelated to topology. The reset makes the assertion about the property rather than about test order.

## Deliberately not changed

- `create_admin_container` is kept, along with its `discover_domains()` loop — that loop is what populates the attributes `_discover_and_register_pages` reads at bootstrap time.
- The worker/broker module still holds its own `CoreContainer`. That is the third tree, and it is #324's to address.

### DIFF
--- a/src/_apps/admin/bootstrap.py
+++ b/src/_apps/admin/bootstrap.py
@@ -46,7 +46,10 @@ _global_exception_handler_registered = False
 
 def bootstrap_admin(fastapi_app: FastAPI) -> None:
     """Bootstrap NiceGUI admin dashboard onto the existing FastAPI app."""
-    admin_container = create_admin_container()
+    # A view over the server's tree, not a second one — see
+    # ``create_admin_container`` for why the core container cannot simply be
+    # injected the way the worker does it.
+    admin_container = create_admin_container(fastapi_app.state.container)
     fastapi_app.state.admin_container = admin_container
 
     # Global safety net (#195): uniformly structured-log any exception that
@@ -66,8 +69,11 @@ def bootstrap_admin(fastapi_app: FastAPI) -> None:
     # admin action can fire. The logger and the repository share the same
     # instance — the logger uses it for writes, the audit-log UI for queries,
     # and the cleanup task for retention deletes.
-    database = admin_container.core_container.database()
-    audit_repo = AdminAuditLogRepository(database)
+    # Pass the *provider*, not a resolved Database. Resolving here froze the
+    # connection into module state, so an override applied after bootstrap
+    # (override_database in the e2e harness) was invisible to every audit write
+    # and query while /v1/* ran on the swapped database.
+    audit_repo = AdminAuditLogRepository(admin_container.core_container.database)
     configure_audit_repository(audit_repo)
     configure_audit_logger(AuditLogger(audit_repo))
     configure_admin_account_use_case_provider(
--- a/src/_apps/admin/di/container.py
+++ b/src/_apps/admin/di/container.py
@@ -1,25 +1,59 @@
-from dependency_injector import containers, providers
+from dependency_injector import containers
 
-from src._core.infrastructure.di.core_container import CoreContainer
-from src._core.infrastructure.discovery import discover_domains, load_domain_container
+from src._core.infrastructure.discovery import discover_domains
 
 
-def create_admin_container() -> containers.DynamicContainer:
-    """Dynamically create the admin DI container.
+def create_admin_container(server_container) -> containers.DynamicContainer:
+    """Create the admin DI container as a view over the server's own tree.
 
-    Same pattern as create_server_container() — auto-discovers all domains.
+    Args:
+        server_container: the server ``DynamicContainer`` from
+            ``app.state.container``. The admin dashboard is mounted onto the
+            server process (``ui.run_with``), so it shares that process's
+            containers rather than building a second set.
+
+    Why this is a view and not its own tree
+    ---------------------------------------
+    This function used to call ``providers.Container(CoreContainer)`` and rebuild
+    every domain container. One process then held **two** of every
+    ``CoreContainer`` Singleton: two async engines and two QueuePools (with
+    ``database_pool_size``/``database_max_overflow`` unset, up to 2 x (5 + 10) =
+    30 connections against a 15-connection budget), two ``HttpClient``s, and two
+    ``ErrorNotifier`` cooldown dicts — making the "per-process" notification
+    cooldown really per-tree. Every optional-infra disabled warning was emitted
+    twice, including the ``notification_client_disabled`` line that
+    ``test_noop_notification_client_disabled_warning_emitted_once`` pins to
+    exactly one occurrence. And ``override_database`` reached only the server
+    tree, so a test that swapped the database got ``/v1/*`` on the swapped one
+    while ``/admin/*`` and every audit write stayed on the real ``DATABASE_*``.
+
+    Injecting the core container the way ``create_worker_container`` does is not
+    available here, and this is the part worth reading before changing it:
+    ``providers.Container(DomainContainer, core_container=...)`` **overrides
+    class-level state** on the domain container. By the time admin mounts, the
+    server has already applied that override, and re-applying it with a
+    core container that is already in use raises
+
+        AttributeError: 'DependenciesContainer' object has no attribute '__self__'
+
+    Only a *brand new* core tree is accepted as the second override — which is
+    the very thing being removed. The worker pattern works because a worker
+    process wires those classes exactly once. So admin reuses the server's
+    domain container providers rather than constructing its own.
+
+    ``discover_domains()`` still drives the loop: the attributes it sets are what
+    ``_discover_and_register_pages`` reads, and NiceGUI page registration happens
+    at bootstrap time. The values stay *providers*, never resolved services, so
+    the ``page_config._service_provider`` indirection is preserved.
     """
     container = containers.DynamicContainer()
-    container.core_container = providers.Container(CoreContainer)
+    container.core_container = server_container.core_container
 
     for domain in discover_domains():
-        domain_container_cls = load_domain_container(domain)
         setattr(
             container,
             f"{domain}_container",
-            providers.Container(
-                domain_container_cls, core_container=container.core_container
-            ),
+            getattr(server_container, f"{domain}_container"),
         )
 
     return container
--- a/src/_core/infrastructure/admin/audit/repository.py
+++ b/src/_core/infrastructure/admin/audit/repository.py
@@ -6,6 +6,7 @@ surface used by ``/admin/audit-log`` and the cleanup task: ``list_filtered``
 detail dialog), and ``delete_older_than`` (retention cleanup).
 """
 
+from collections.abc import Callable
 from datetime import UTC, datetime
 
 from sqlalchemy import delete, func, select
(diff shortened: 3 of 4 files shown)


Mr-DooSun:
Cross review (codex, read-only) on `c0fd930`: **PASS**, no findings, correctness and security.

It verified each concern I raised rather than accepting the diff:

- **Aliasing does not break what admin needed from a separate tree.** Page registration completes before `ui.run_with()`, and `_service_provider`, the admin auth provider, the permission registry, and the startup seed all resolve lazily through the same `admin_identity_container`.
- **`isinstance(source, Database)` is the right discrimination.** Real `Database` instances are not callable, and the type test also handles a hypothetical callable subclass correctly — which is why it replaced `callable(source)`.
- **No intra-method database swap.** All four `async with self._database.session()` sites resolve the property once to obtain the context manager, so two accesses inside one method cannot get different databases. Re-resolution costs a Singleton lookup.
- **Nothing load-bearing was isolated per tree.** NiceGUI mounts into the same process and event loop; the shared async engine creates a fresh session per call, and the shared `HttpClient` / `ErrorNotifier` cooldown are intended process-wide behaviour — that was the bug, not a feature.
- **The `user_repository.reset()` reasoning holds.** A Singleton repository built before the override keeps the old database, and because server and admin alias one provider, a single reset covers both. Teardown order (release the override, then reset) preserves isolation.

It did not run the full suite, being read-only; `1597 passed, 42 skipped, 0 failed` locally.

Issue #368, opened 2026-08-11
Admin UI/UX: redesign the login and dashboard composition (a token swap could not fix "it doesn't look changed")


Mr-DooSun:
## Summary

#365 replaced the admin palette and shape tokens, and that shipped (PRs #366, #367). Looking at the result, the honest verdict is that **it does not read as a new design.** This issue is the follow-up: redesign the *composition* of the two screens that form the entire first impression.

## Why a token swap could not have fixed this

Tokens change colour and shape. They cannot change what is on the screen or how it is arranged — and the arrangement is what a person remembers. Both first-impression screens are composed identically before and after #365:

| screen | composition (unchanged by #365) |
|---|---|
| `/admin/login` | one 360px card: icon → brand name → the literal string `ADMIN` → 2 inputs → 1 button |
| `/admin/` (dashboard) | 4 stat cards → 1 bar chart → 1 table |

These two carry the whole first impression: login is unavoidable, and the dashboard is where login lands. List and detail pages are shaped by AG Grid, so there is little composition to redesign there — which is why they are **out of scope** here.

## The dashboard's real problem is not density, it is that it answers nothing

On a fresh quickstart the dashboard reads: `AI Usage 0` · `Docs 0` · `User 0` · `Audit Events 12`, a single-category bar chart whose only category is `LOGIN`, and a table of consecutive `LOGIN` rows. It is a list of counts.

An operator dashboard earns its place by answering "what should I look at right now?" A prettier stat card does not fix a screen that has no question to answer. **This is the part of the issue that cannot be started without a decision** — see Open Questions.

The login screen has no such ambiguity: its job is self-evident and its scope is small, so it can proceed independently.

### Login (`src/_apps/admin/pages/login.py`) — ready to start
- [ ] Reconsider the composition, not the colours. Current shape is a single centred card on a flat backdrop with no supporting content
- [ ] The literal `ADMIN` label under the brand name — decide whether it earns its line
- [ ] Error/failed-login state: check what it looks like today and whether it reads as recoverable
- [ ] `render_dark_mode_toggle()` currently floats alone in the top-right corner of an otherwise empty page
- [ ] Keep the pre-auth dark-mode toggle working (it is why login calls the shared toggle rather than the shell)

### Dashboard (`src/_apps/admin/pages/dashboard.py`) — blocked on Open Questions
- [ ] Decide what questions the dashboard answers (see below)
- [ ] Recompose around those answers rather than around available counts
[...]

PR #372, opened 2026-08-12
feat(admin): replace what the dashboard answers, with onboarding as a real state

Mr-DooSun:
Second of two PRs for #368. Based on `feat/368-base-repo-daily-count` (**#371**) — merge that first; this branch contains it, so the diff here is only the dashboard work.

## What the dashboard answers now

| | before | after |
|---|---|---|
| sections | stat cards · audit activity chart · recent-audit table | what is wired up · agent calls + failure rate · new records (7d) |
| zero-data state | the same page with every number at 0 | a distinct onboarding screen |
| audit reads per load | 1 | 0 |

The audit sections are gone **and so is the collection behind them**. `/admin/audit-log` is a strict superset with filtering and pagination, so the landing page was reading audit data to render a worse copy of a page that already existed.

## Zero data is the default, so it gets a real screen

Every count is 0 on a fresh install — that is what every OSS adopter opens first and what a demo recording captures. Charting zeros there looks broken. That path now shows what is configured plus the next action: *"Run make demo"*, *"LLM is running a stub — set LLM_PROVIDER + LLM_MODEL"*.

`has_any_data` deliberately ignores infrastructure: every install has infra to report, so counting it would mean the onboarding view never appears. Unavailable sources (`None`) are not evidence of data either, so a broken backend does not skip past onboarding into an empty dashboard.

## Two decisions worth arguing with

**Failures are `total - ok`, not a sum of failure states.** `UsageStatus` is `Literal["ok", "error", "timeout", "rate_limited"]`. Summing the three known failures means a fourth state added later is silently counted as *success*. Subtracting from the total cannot drift that way.

**`failure_rate` is `None`, not `0.0`, when there are no calls.** A rate needs a denominator, and a healthy 0% for a system that has served nothing is a false reassurance. It renders as an em dash.

## Reachability, not new coupling

Both new sources are capability-probed off the service, matching how `count_datas` is already reached rather than importing `AiUsageService`:

- a fork that drops `ai_usage` loses the panel, not the page
- a domain whose model has no temporal column degrades its own section (the repository raises a curated 400; the facade catches it per-section)
- `visible_configs` is already permission-filtered, which is what keeps the AI read scoped — an operator without the permission never causes it

## Three layout defects were mine, and only the screenshot pass found them

None was reachable by a unit test:

1. **One chart per domain reproduced the defect #369 had just removed.** Each chart is a fixed 260px, so two domains pushed the page to **1740px** with two nearly-empty plots. Replaced with one aggregate series — the per-domain totals are already on the stat cards, and what the chart adds is shape over time.
2. **`c.empty_state` is a whole-page placeholder** (48px vertical padding, centred). Using it for an inline "no agent calls" note spent ~250px on one sentence. Muted labels instead.
3. **Step cards sized to their own text**, so the onboarding column had ragged right edges that read as a layout bug.

Page height across the three fixes: **1740 → 1255px**.

## Verification

| gate | result |
|---|---|
| `make check-core` (SQLite) | 1862 passed, 29 skipped |
| `make test-pg` (real PostgreSQL) | 1989 passed, 43 skipped |
| `make check-minimal` | 7 passed |
| `pyright` | 0 errors |

Visual: both light and dark, on the **onboarding path** (fresh database) and the **populated path** (after `make demo` + `make demo-rag`). The two paths are separate branches in the page, so exercising only one would have proven half of it.

## Security note — please read with the operational-status facade

The enable/disable predicates in `core_container` key off credentials (`dynamodb_access_key`, `s3vectors_access_key`, `notification_webhook_url`), so the code deciding what the panel shows reads secrets by construction. Reporting a state derived from a secret is fine; reporting the secret is not.

Enforced mechanically rather than by review judgement: a value renders only if its settings field is on an allowlist, `_detail` raises on anything else, and a test populates every credential-shaped field with a sentinel and searches the rendered output for it. A second test guards the allowlist itself against credential-shaped names, and a third compares the panel's verdict against each container selector so the duplicated predicates cannot silently diverge.

`STUB` and `DISABLED` are kept distinct because ADR 042 gives optional infra two different disabled behaviours, and conflating them hides the one that bites: a stub *appears* to work.

## Not done here

`/security-review` and `/review-architecture` from the execution packet have **not** been run — the packet marks the security review L3 (sensitive-data item applies), so it is the next step rather than something to self-certify in a PR description.

### DIFF
--- a/src/_apps/admin/dashboard_metrics.py
+++ b/src/_apps/admin/dashboard_metrics.py
@@ -1,45 +1,43 @@
-"""Dashboard read facade for the admin landing page.
+"""What the admin landing page reads (#193, reworked in #368).
 
-Isolates **all** data gathering for ``/admin/`` so the page file
-(``pages/dashboard.py``) stays a thin renderer — a core admin landing page must
-not embed cross-domain orchestration, dynamic service probing, or error
-isolation inline.
+A façade that **never raises**: every source is isolated so one failing backend
+degrades its own section instead of 500-ing the page. Sources that are absent
+report ``None``, which the page renders as "Unavailable" — distinguishable from a
+real zero, because "the count is 0" and "the count could not be read" are
+different facts and collapsing them hides outages.
 
-Design invariants:
-
-* **Never raises into the caller.** Each metric source is fetched in its own
-  best-effort block; one failing source degrades *that* section to an explicit
-  "unavailable" state (``count=None`` / ``audit_recent=None``) and is recorded
-  via structlog. The dashboard as a whole never breaks because one service is
-  down.
-* **No raw exception text leaves this module.** Failures are logged with
-  ``error_type`` only; the UI renders a neutral "Unavailable" marker. This keeps
-  the no-``str(exc)``-leak invariant (#195) intact end to end.
-* **CRUD contract stays minimal.** ``count_datas`` is implemented by every
-  ``BaseService`` but is intentionally *not* on ``AdminCrudServiceProtocol``.
-  The single dynamic ``getattr`` probe is quarantined here rather than widening
-  the shared CRUD protocol just to feed a dashboard.
+What the dashboard answers, and what it stopped answering
+---------------------------------------------------------
+#368 replaced the questions. The page now reports **what is wired up**, **AI
+usage and its failure rate**, and **per-domain growth**. The audit-derived
+sections it used to carry (an activity chart and a recent-activity table) are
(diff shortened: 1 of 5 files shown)


Mr-DooSun:
Closed automatically when its base branch `feat/368-base-repo-daily-count` was deleted on merging #371 — a PR whose base branch is gone cannot be reopened. Continued as #373 with the same commits rebased onto `main`.

Issue #375, opened 2026-08-12
Manual mypy hook cannot type-check anything: duplicate completion_gate module across .claude and .antigravity


Mr-DooSun:
## Summary

The manual mypy stage cannot type-check anything. It aborts during module resolution:

```
$ uv run pre-commit run --hook-stage manual mypy --all-files
.claude/hooks/completion_gate.py: error: Duplicate module named "completion_gate"
  (also at ".antigravity/hooks/completion_gate.py")
Found 1 error in 1 file (errors prevented further checking)
```

`errors prevented further checking` is the important part: mypy stops before inspecting a single file under `src/`, so the hook reports one error and zero real findings.

## Cause

`.claude/hooks/completion_gate.py` and `.antigravity/hooks/completion_gate.py` are separate files with the same basename and no package boundary between them, so both map to the top-level module name `completion_gate`. The hook's `exclude` covers `migrations/`, `tests/`, `.*test\.py$`, `run_.*\.py$` and `force_.*\.py$` — not the harness hook directories.

Harness parity is why the pair exists (ADR 045 / the Phase 5 shared governor package), so the duplicate name is intentional; only mypy's view of it is wrong.

## Why it went unnoticed

CI's `typecheck` job runs **pyright**, not mypy, scoped by the `[tool.pyright]` allow-list. So CI is genuinely green while the mypy hook has been unable to run. Nothing depends on mypy today, which is exactly why this could sit unnoticed — but it also means the hook is dead weight that looks like coverage.

## Possible directions

- Add `.claude/hooks/`, `.antigravity/hooks/` (and any sibling harness dirs) to the hook's `exclude`
- Or pass `--explicit-package-bases` / set `MYPYPATH` so the two paths resolve distinctly
- Or retire the mypy hook and state that pyright is the type-check authority, since that is already the CI reality

The third option is worth considering on its own merits: two type checkers with different scopes, one of which cannot run, is more confusing than one.

Found while measuring the blast radius of a protocol change in #368 / #371.

Mr-DooSun:
## Measurement — the "one line of `exclude`" framing in this issue was wrong

I filed this suggesting the fix might be a single `exclude` addition. I measured it, and it is not.

| | |
|---|---|
| files the hook tries to check | **681** |
| duplicate basenames between harness dirs | **8 pairs**, not one |
| files after excluding harness dirs | 641 |
| **real findings once mypy can run** | **46 errors in 21 files** |

[...]

Issue #374, opened 2026-08-12
make test-pg fails against a non-empty PostgreSQL: drop_all cannot drop "user" while refresh_token's FK exists


Mr-DooSun:
## Summary

`make test-pg` only works against a database that has no tables yet. If the target PostgreSQL already holds the schema — from a previous run, from `make dev`, or from an Alembic upgrade — the session fixture dies during **setup**:

```
asyncpg.exceptions.DependentObjectsStillExistError:
  cannot drop table "user" because other objects depend on it
DETAIL:  constraint refresh_token_user_id_fkey on table refresh_token depends on table "user"
```

Every test in the file errors, so it reads like a broken change rather than a broken fixture.

## Cause

`tests/conftest.py::test_db` runs `Base.metadata.drop_all` before `create_all`. `drop_all` emits plain `DROP TABLE` in dependency order derived from the metadata, but `refresh_token.user_id` has a FK to `user`, and PostgreSQL refuses the drop while the dependent constraint exists. SQLite does not enforce this, which is why CI is green: CI runs SQLite, and `make test-pg` is normally pointed at a container created fresh, where `drop_all` is a no-op.

## Reproduce

```bash
docker compose -f docker-compose.local.yml up -d postgres
make test-pg          # first run: passes, leaves the schema behind
make test-pg          # second run: every test in the file ERRORs at setup
```

Confirmed against an **existing** test (`TestBulkInsertAcrossEngines::test_empty_bulk_insert_is_a_noop`), so this is not specific to any recent change.

Workaround today:

```bash
docker exec fastapi-agent-blueprint-postgres \
  psql -U postgres -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"
```

## Why it matters

`make test-pg` is the documented way to verify the production engine, and ADR 058 leans on it — the base-repository contract tests exist precisely because SQLite and PostgreSQL diverge. A verification path that silently requires a pristine database is one people stop trusting, and the failure mode points at the wrong thing.

## Possible directions

- `drop_all` after a `CASCADE`-aware teardown, or drop the schema rather than the tables
- Create the schema once per session and `TRUNCATE` between runs instead of drop/create
- Keep drop/create but make the failure explain itself ("target database is not empty; run …")

Found while adding engine-parity tests in #368 / #371.

PR #376, opened 2026-08-12
docs: sync shared references to the reworked admin dashboard

Mr-DooSun:
Closes the drift that both review gates on #368 reported. Doc-only.

## The one that matters: a rule that did not exist

`security-checklist.md` §2 gains: **a dashboard section must not exceed the permission scope of its subject, and the read must be gated rather than only the render.**

The principle was already practised. The pre-#368 dashboard gated its audit sections on the `audit_log` permission and passed `include_audit=False` so the audit repository was never touched without it — with a comment saying *"the landing page must not surface audit data more broadly than the dedicated audit-log page"*.

But it lived **only in that comment**, on the code being deleted. So when #373 replaced those sections with an infrastructure panel that had no gate, nothing failed: no rule to cite, no test to break. `/security-review` caught it by reading `require_auth_allowlisted()` and noticing it authenticates without checking permissions — not by running a check. The panel was visible to an admin holding zero grants for exactly one review round.

That is the failure mode worth fixing in the checklist rather than only in the code: *"Error notification: stub"* tells the holder of a low-privilege admin account that failures raise no alert.

## Builder catalog corrections

`c.empty_state` is now documented as a **full-panel** placeholder — centred, large icon, `48px` vertical padding — explicitly not for an inline note. The row previously read only "Centered empty placeholder; add the message inside", which is what let me use it for a one-line remark inside a populated section at a cost of ~250px.

The `c.data_grid` row already stated the caller-bounded requirement for `auto_height`, so it is **left alone** rather than padded for the sake of touching it.

## Rules files

- `project-overview.md` — `DailyCount` added to Key Value Objects, with the "days with no rows are absent, not zero-filled" policy note
- `architecture-conventions.md` — records the pairing rule this work uncovered: an aggregate read primitive that callers need must exist on `BaseService` as well as the repository, because callers hold a service (the dashboard resolves `config._get_service()`, never a repository)
- `commands.md` — the `make test-pg` clean-database caveat plus the reset command. Without it a stale schema makes **every** test in the file error at setup and look like a broken change (#374)
- `project-status.md` — a row for #368 recording the two measured dialect traps, the measured protocol blast radius, and the security finding

## Verification

- `pre-commit run --all-files` — exit 0, including the Tier-1 English-only policy
- `project-status.md` table at **11 rows**, under the 15-row archival trigger; no multi-line cells
- Exactly one `Last synced` line per rules file, all dated 2026-08-12 — checked because a previous sync in this series accidentally left two in `project-overview.md`

## Related issues filed rather than fixed

- #374 — `make test-pg` fails against a non-empty PostgreSQL (`drop_all` vs the `refresh_token → user` FK). Reproduced with a pre-existing test
- #375 — the manual mypy hook cannot type-check anything (duplicate `completion_gate` module across `.claude/hooks/` and `.antigravity/hooks/`)

## Governor Footer

- trigger: yes
- reviewer: self-structured
- rounds: 1
- r-points-fixed: 1
- r-points-deferred: 0
- r-points-rejected: 0
- pr-scope-notes: Doc-only sync closing drift from /review-architecture and /security-review on #368. R1 = security-checklist.md had no rule for within-page permission scoping, which is why an ungated infrastructure panel passed review; Fixed by adding the rule. Cross-tool Codex review not attempted: codex exec timed out with no final answer twice earlier in this session on codex-cli 0.144.4, so self-structured per AGENTS.md Independent Review Trigger. No ADR consequence added — the new checklist item is a review rule, not a durable architecture constraint, and the admin design-system entries are catalog corrections.
- final-verdict: merge-ready
- links: https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/376, n/a


Mr-DooSun:
**Correction to this PR's Governor Footer.**

The footer records `reviewer: self-structured` with the reason "codex exec timed out with no final answer twice on codex-cli 0.144.4". **That attribution was wrong.**

The real cause was that `codex exec` blocks on `Reading additional input from stdin...` — it waits for stdin EOF even when the prompt is passed as an argument, and the calling shell left stdin open. `codex exec --sandbox read-only ... < /dev/null` returns normally; a cross-review run that way completed in a few minutes and produced a full result.

So the fallback to `self-structured` was based on a misdiagnosed tool failure, not an unavailable tool. The self-structured review itself stands on its own — the R-points and verification in this PR were real — but a genuine cross-tool review was available and I did not obtain one.

Correcting here rather than rewriting merged history. Detail and the reproduction are in #375.

PR #377, opened 2026-08-12
fix(tests): load every model before the schema fixture runs

### LINKED ISSUE #374: make test-pg fails against a non-empty PostgreSQL: drop_all cannot drop "user" while refresh_token's FK exists

Mr-DooSun:
Closes #374.

## The bug

`tests/conftest.py` never imported the models. So `Base.metadata` held only whatever the collected test modules happened to import — measured as `['user']` for the entire `persistence/rdb` directory.

On PostgreSQL that made `drop_all` order-dependent. Metadata unaware that `refresh_token.user_id` references `user` tried to drop `user` first, and PostgreSQL refused while `refresh_token_user_id_fkey` existed. Every test in the file errored at **setup**, which reads as a broken change rather than a partial-metadata problem.

Two things hid it:

- the full-suite run imported enough models **by accident**
- SQLite enforces no FKs and starts from a fresh in-memory database

So it only surfaced when someone ran a subset against a PostgreSQL that already held the schema — from `make dev`, an `alembic upgrade`, or an earlier run.

## The fix

Call the existing Alembic `load_models()` helper at conftest import time. Reusing it rather than re-listing model paths keeps one source of truth for where models live. It imports only `src/*/infrastructure/database/models/**` plus the explicit `_core` packages, so it pulls in no optional-extra dependency — `make check-minimal` still passes.

## Causality was established, not assumed — and it took three attempts

My first two measurements contradicted each other: the same command appeared to pass with and without the fix. The cause was my own method — a previous `drop_all` had already destroyed the schema I was measuring against, and `alembic upgrade head` silently did nothing because `alembic_version` still said *head*. So I was testing against an empty database and proving nothing.

Verified from one clean state per measurement: reset the schema, `alembic upgrade head`, confirm `refresh_token_user_id_fkey` exists, then run the same single test.

| condition | result |
|---|---|
| without `load_models()` | **ERROR at setup** (`DBAPIError`) |
| with `load_models()` | 1 passed |

The mechanism was also confirmed in isolation: with metadata built from only the user model, a standalone `drop_all` against an FK-bearing schema raises; with `load_models()` it succeeds.

`make test-pg` against a pre-existing schema — the scenario the issue reports — now passes in full: **1997 passed**.

## Tests

`test_metadata_completeness.py` pins three things: every expected table is registered, both ends of each cross-domain FK are visible to metadata, and the derived **drop order** places dependants before their targets (the property `drop_all` actually relies on).

Its docstring records why running that file alone is the meaningful check: in isolation nothing else imports the models, so removing the `load_models()` call fails there, while a full-suite run would pass either way — which is precisely why the defect survived.

## Also

`commands.md` loses the manual-reset caveat added earlier today. It is now false, and a stale workaround is worse than none.

## Verification

| gate | result |
|---|---|
| `make test-pg` **against a non-empty schema** | 1997 passed, 43 skipped |
| `make check-core` (SQLite) | 1870 passed, 29 skipped |
| `make check-minimal` | 7 passed |
| `pyright` | 0 errors |
| `pre-commit run --all-files` | exit 0 |

## Governor Footer

- trigger: yes
- reviewer: self-structured
- rounds: 1
- r-points-fixed: 1
- r-points-deferred: 0
- r-points-rejected: 0
- pr-scope-notes: Governor surface is one line in .claude/rules/commands.md removing a now-false caveat. R1 = my first two before/after measurements contradicted each other because I measured against a schema a previous drop_all had already destroyed; Fixed by re-measuring from one clean state per run and confirming the FK exists first. Cross-tool Codex review not attempted: codex exec timed out with no final answer twice earlier in this session on codex-cli 0.144.4, so self-structured per AGENTS.md Independent Review Trigger. No ADR consequence added — this is a test-fixture correctness fix, not a durable architecture constraint.
- final-verdict: merge-ready
- links: https://github.com/Mr-DooSun/fastapi-agent-blueprint/pull/377, n/a

### DIFF
--- a/.claude/rules/commands.md
+++ b/.claude/rules/commands.md
@@ -1,6 +1,6 @@
 # Suggested Commands
 
-> Last synced: 2026-08-12 via #368 (added the `make test-pg` clean-database caveat — see #374; no new commands). Prior: 2026-08-11 via #365/PR #366 (Admin Dashboard — the UI theming block now names the neutral-mono theme instead of the Toss-style one, and records that the `--q-*` brand group must be emitted under `body` with `!important`, because NiceGUI writes that palette as an inline body style and a `:root` declaration is inert — the pre-#365 state, in which the whole brand half of the palette never applied). Prior: 2026-08-01 via #286/PR #313 + #315/PR #319 (Error Notification — severity channel routing section and example added; `:159` corrected, it had contradicted the Coverage bullet in the same block since #310). Prior: 2026-07-28 via #310 (Error Notification — dispatch extended to Taskiq worker task failures; the server-only scope note from PR #311 is superseded). Prior: 2026-07-27 via #307/PR #311 (Error Notification section — runbook pointer, server-only dispatch scope, first-dispatch disabled warning). Prior: 2026-07-23 via #17/PR #304 (added the Error Notification section — `NOTIFICATION_*` env vars for Slack/Discord webhook alerts fired from the exception handlers). Prior: 2026-07-20 via ADR 056 (added `tools/check_migration_safety.py` — advisory unsafe-DDL scan for zero-downtime migrations — to Architecture Verification + DB Migrations). Prior: 2026-07-20 via #293 (added `make perf-test` — Locust performance-test harness — to the Test section).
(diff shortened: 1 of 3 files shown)


Mr-DooSun:
**Correction to this PR's Governor Footer.**

The footer records `reviewer: self-structured` with the reason "codex exec timed out with no final answer twice on codex-cli 0.144.4". **That attribution was wrong.**

The real cause was that `codex exec` blocks on `Reading additional input from stdin...` — it waits for stdin EOF even when the prompt is passed as an argument, and the calling shell left stdin open. `codex exec --sandbox read-only ... < /dev/null` returns normally; a cross-review run that way completed in a few minutes and produced a full result.

So the fallback to `self-structured` was based on a misdiagnosed tool failure, not an unavailable tool. The self-structured review itself stands on its own — the R-points and verification in this PR were real — but a genuine cross-tool review was available and I did not obtain one.

Correcting here rather than rewriting merged history. Detail and the reproduction are in #375.

